PR #6, opened 2020-04-09
feature: formRPartA resource in trainee-forms

hihilary:
1. API for get / save formRPartA, including entity, Dto, etc.
2. unit tests

TISNEW-3946


hihilary:
These duplication seems to due to same codes in Dto and entity, which shouldn't matter.

PR #38, opened 2020-09-21
feat: Use the trainee's TIS ID from token

Judge40:
Currently the trainee ID is provided as part of the request, but this
would potentially allow users to access data that is not theirs. Instead
the authentication token should be used to retrieve the logged in user's
TIS ID directly.

TISNEW-3833

### DIFF
--- a/src/main/java/uk/nhs/hee/tis/trainee/forms/api/util/AuthTokenUtil.java
+++ b/src/main/java/uk/nhs/hee/tis/trainee/forms/api/util/AuthTokenUtil.java
@@ -1,0 +1,69 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright 2020 Crown Copyright (Health Education England)
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of this software and
+ * associated documentation files (the "Software"), to deal in the Software without restriction,
+ * including without limitation the rights to use, copy, modify, merge, publish, distribute,
+ * sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all copies or
+ * substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT
+ * NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+ * DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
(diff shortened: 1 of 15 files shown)


Judge40:
Created as draft due to API Gateway changes that are required for this to be merged successfully.

uzairbaig (approved):

src/main/java/uk/nhs/hee/tis/trainee/forms/api/util/AuthTokenUtil.java
```diff
@@ -0,0 +1,88 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.Base64;
+import java.util.Map;
+import java.util.Optional;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+
+@Slf4j
+public class AuthTokenUtil {
```
paulhoang-hee, line 35: My preference would be to turn this into a spring bean and remove the static keywords

src/main/java/uk/nhs/hee/tis/trainee/forms/api/util/AuthTokenUtil.java
```diff
@@ -0,0 +1,88 @@
...
+
+  private static final String TIS_ID_ATTRIBUTE = "custom:tisId";
+
+  private static final ObjectMapper mapper = new ObjectMapper();
+
+  private AuthTokenUtil() {
```
paulhoang-hee, line 41: if we're dealing with tokens, should we not validate them against the secret?

src/main/java/uk/nhs/hee/tis/trainee/forms/api/util/AuthTokenUtil.java
Judge40: You're probably right, old habits die hard.

src/main/java/uk/nhs/hee/tis/trainee/forms/api/util/AuthTokenUtil.java
Judge40: Wasn't sure about this when I was doing it to be honest. The expectation is that the service itself will always be private and public access is only via API Gateway. Given the API Gateway will authorize the token I've assumed, perhaps naïvely, that the token can be trusted within this service to keep things fairly simple.

It does rely on the service being kept private and only ever accessible through the API Gateway, I suppose there is room for error there.

DorisWongNHS (approved):

src/main/java/uk/nhs/hee/tis/trainee/forms/api/util/AuthTokenUtil.java
paulhoang-hee: if the api gateway does the jwt validation then I don't see any issue with this. last thing you want is to duplicate this validation code everywhere in other services - yay for shared modules

src/main/java/uk/nhs/hee/tis/trainee/forms/api/util/AuthTokenUtil.java
```diff
@@ -0,0 +1,88 @@
...
+
+  /**
+   * Verify that the token contains a valid trainee TIS ID and that it matches the expected ID.
+   *
+   * @param requestedTraineeTisId The ID to validate against.
+   * @param token                 The token to use.
+   * @param <T>                   The type of the response body.
+   * @return A ResponseEntity with a suitable error status, or empty if valid.
+   */
+  public static <T> Optional<ResponseEntity<T>> verifyTraineeTisId(String requestedTraineeTisId,
```
paulhoang-hee, line 69: I'd be tempted to remove the HTTP related stuff here and just return a boolean here to state if its valid

paulhoang-hee (approved):
good to go

PR #60, opened 2020-10-26
refactor: define an abstract form and cloud repository

JosephKelly:
This change should fix duplication code smell and make more testable.

TISNEW-5376: Show unsubmitted forms


Judge40 (approved):
Really nice refactoring, good use of generics to create the abstract.

PR #186, opened 2021-04-08
refactor: Fix code smells

Judge40:
TIS21-1397

### DIFF
--- a/src/main/java/uk/nhs/hee/tis/trainee/forms/api/FormRPartAResource.java
+++ b/src/main/java/uk/nhs/hee/tis/trainee/forms/api/FormRPartAResource.java
@@ -155,7 +155,6 @@ public class FormRPartAResource {
   @GetMapping("/formr-parta/{id}")
   public ResponseEntity<FormRPartADto> getFormRPartAsById(@PathVariable String id,
       @RequestHeader(HttpHeaders.AUTHORIZATION) String token) {
-    // TODO: verify the formR PartA for the id belongs to the trainee
     log.debug("FormR-PartA by id {}", id);
 
     String traineeTisId;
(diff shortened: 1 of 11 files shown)


src/main/java/uk/nhs/hee/tis/trainee/forms/api/FormRPartAResource.java
```diff
@@ -155,7 +155,6 @@ public FormRPartAResource(FormRPartAService service, FormRPartAValidator validat
   @GetMapping("/formr-parta/{id}")
   public ResponseEntity<FormRPartADto> getFormRPartAsById(@PathVariable String id,
       @RequestHeader(HttpHeaders.AUTHORIZATION) String token) {
-    // TODO: verify the formR PartA for the id belongs to the trainee
```
ReubenRobertsHEE, line 158: Is this TODO resolved somewhere or no longer relevant? I vaguely recall some discussions around securing access to the saved forms, but I'm not sure if that is related.

ReubenRobertsHEE: Otherwise it all looks fine to me.

Judge40: The service takes the trainee ID as a parameter (retrieved from the token) which is then used to build the path to the form. If the trainee ID isn't correct for the form ID then it won't exist on the built path.

As far as I could see, the concerns raised in the TODO are covered already.

Judge40: The service takes the trainee ID as a parameter (retrieved from the token) which is then used to build the path to the form. If the trainee ID isn't correct for the form ID then it won't exist on the built path.

As far as I could see, the concerns raised in the TODO are covered already.

ReubenRobertsHEE (approved):
LGTM

PR #192, opened 2021-04-15
feat: Add Mongock insert of COVID form switch

Judge40:
The form switch for COVID needs to be inserted if it is missing, use
mongock change logs to add the entry to the database.

TIS21-1452

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,7 +10,7 @@ plugins {
 }
 
 group = "uk.nhs.hee.tis.trainee"
-version = "0.2.1"
+version = "0.3.0"
 sourceCompatibility = "11"
 
 configurations {
@@ -44,6 +44,11 @@ dependencies {
 
   // Sentry reporting
   implementation "io.sentry:sentry-spring-boot-starter:4.3.0"
+
+  // Mongock
+  ext.mongockVersion = "4.3.8"
+  implementation "com.github.cloudyrock.mongock:mongock-spring-v5:${mongockVersion}"
+  implementation "com.github.cloudyrock.mongock:mongodb-springdata-v3-driver:${mongockVersion}"
 }
 
 checkstyle {
--- a/src/main/java/uk/nhs/hee/tis/trainee/forms/changelog/FormSwitchChangeLog.java
+++ b/src/main/java/uk/nhs/hee/tis/trainee/forms/changelog/FormSwitchChangeLog.java
@@ -1,7 +1,6 @@
 /*
  * The MIT License (MIT)
- *
- * Copyright 2020 Crown Copyright (Health Education England)
+ * Copyright 2021 Crown Copyright (Health Education England)
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy of this software and
  * associated documentation files (the "Software"), to deal in the Software without restriction,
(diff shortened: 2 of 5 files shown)


Judge40:
@Health-Education-England/full-stack I'm looking for some opinions on the overall approach to the form switches, with some specific points that need answering.

 1. Should the change set have `alwaysRun = true`? This will set the value to whatever is currently on the changeset every time the service is deployed, so if a manual change was made in the database (or it was deleted) it would automatically get re-set.
 2. Is there even an advantage to storing the switch values in the DB? If we only ever control it via the changeset is there an advantage over just having a hardcoded value in the `FormSwitch` DTO? (especially if alwaysRun is true)

jayanta2018:
> @Health-Education-England/full-stack I'm looking for some opinions on the overall approach to the form switches, with some specific points that need answering.
> 
> 1. Should the change set have `alwaysRun = true`? This will set the value to whatever is currently on the changeset every time the service is deployed, so if a manual change was made in the database (or it was deleted) it would automatically get re-set.
> 2. Is there even an advantage to storing the switch values in the DB? If we only ever control it via the changeset is there an advantage over just having a hardcoded value in the `FormSwitch` DTO? (especially if alwaysRun is true)

@Judge40 
I have done some research on MongoCk. It is "java MongoDB tool for tracking, managing and applying database schema changes across all your environments based on a coding approach". Thanks for implementing a new stuff.

As Covid situation is uncertain, I think we should go for an option which is more flexible.

jayanta2018 (approved):

JosephKelly:
IMO... this has always looked a lot like feature flags/toggles/switches to me.

Thoughts about the larger question that "spring" to mind:
- If it is reasonably static/"once or twice" over an extended time we'll change this then an application property / changelog sounds good.  I think yml is a bit friendlier than the 
- I see the DB as providing one particular method of 'hot swap' updates. I'd be happy using a)cloud config (for something that doesn't change too often or b)a full 'feature flag' service for something that will change more often or we want to experiment with.
  - A DB makes sense if we are going to have an 'administration' type view.

On the properties for Mongock:
- I can't think why we'd want to reset this every time a container starts/restarts so don't think we'd set `alwaysRun` for this but realise I have missed a fair bit of the conversation for this.

JosephKelly:
I really like the use Mongock for managing changes that we need to make though!  Presumably we would use it to manage indexes and other database operations.

Judge40:
> * I can't think why we'd want to reset this every time a container starts/restarts so don't think we'd set `alwaysRun` for this but realise I have missed a fair bit of the conversation for this.

You haven't missed anything, this PR is literally the only conversation 😬 

The idea behind setting `alwaysRun = true` was that we could change the value in the ChangeSet to modify the value, if we don't always run it then we need an additional change set for every modification of the flag (there's likely to only be one in this case though).

I agree it does feel pretty much like a feature flag, I'm open to reworking this based on any input/advice on the best approach to take.

Judge40:
> I really like the use Mongock for managing changes that we need to make though! Presumably we would use it to manage indexes and other database operations.

We have another technique for adding indexes, I'm not sure whether Mongock could do that but I'll find out!

Judge40:
Using a feature flag technique instead, see #195

PR #195, opened 2021-04-16
feat: Add COVID feature flag

Judge40:
Create a feature flag for the covid declaration as a replacement for the
form switch functionality.

TIS21-1452


JosephKelly (approved):
Really nice simple way of doing it!  I like it.  Means changes can be made really rapidly via cloud provider interfaces :+1:

john12321 (approved):

PR #203, opened 2021-04-28
fix: Resolve race condition on form validation

Judge40:
The form validation is intended to block having multiple drafts of a
form type, however a race condition exists where multiple draft forms
can be validated before either of them are persisted. This means that
both forms pass validation and then are both stored, leaving two drafts.

Remove the manual validation as there is no suitable event that can be
used to avoid the race condition while using this approach.

Add an index to the MongoDB collection to ensure that each trainee can
only have a single draft form for a given type.
Add an exception handler to return a sensible HTTP response for this
scenario.

Remove the specified naming for FormRPartA and FormRPartB collections,
as the given naming does not match the default casing. Using the default
naming convention avoids having to specify the given name in multiple
places of the codebase.

TIS21-1489


src/test/java/uk/nhs/hee/tis/trainee/forms/api/FormRPartAResourceTest.java
```diff
@@ -159,15 +154,7 @@ void postShouldNotCreateFormWhenFormExists() throws Exception {
...
-    dto.setId(null);
-
-    BeanPropertyBindingResult bindingResult = new BeanPropertyBindingResult(dto, "formDto");
-    bindingResult.addError(new FieldError("formDto", "formField", "Form field not valid."));
-
-    Method method = validator.getClass().getMethod("validate", FormRPartADto.class);
-    Exception exception = new MethodArgumentNotValidException(new MethodParameter(method, 0),
-        bindingResult);
-    doThrow(exception).when(validator).validate(dto);
+    when(service.save(dto)).thenThrow(MongoWriteException.class);
```
JosephKelly, line 157: I don't quite get the verification here... we are calling `verifyNoInteractions(service)` which I would have thought means `service.save()` is never invoked.  Given the `postShouldCreate...` succeed, I can't see it being an obvious cause.

Perhaps it's worth adding an `Expect` for the message?

JosephKelly: Ah, I've seen that the test is doing something a bit different now... it is using the validation in the Resource.  That makes this line redundant.

src/test/java/uk/nhs/hee/tis/trainee/forms/api/FormRPartBResourceTest.java
```diff
@@ -159,15 +154,9 @@ void postShouldNotCreateFormWhenFormExists() throws Exception {
...
-    BeanPropertyBindingResult bindingResult = new BeanPropertyBindingResult(dto, "formDto");
-    bindingResult.addError(new FieldError("formDto", "formField", "Form field not valid."));
-
-    Method method = validator.getClass().getMethod("validate", FormRPartBDto.class);
-    Exception exception = new MethodArgumentNotValidException(new MethodParameter(method, 0),
-        bindingResult);
-    doThrow(exception).when(validator).validate(dto);
+    WriteError writeError = new WriteError(1, "exceptionMessage", new BsonDocument());
+    MongoWriteException exception = new MongoWriteException(writeError, null);
+    when(service.save(dto)).thenThrow(exception);
```
JosephKelly, line 159: As per test for PartA resource

JosephKelly (changes requested):
Misc thoughts:
I wondered if the ExceptionHandler would be more maintainable or too broad if it looked an `errorCategory() == ErrorCategory.DUPLICATE_KEY` rather than `errorCode() == 11000` https://mongodb.github.io/mongo-java-driver/ {not sure of version} /javadoc/com/mongodb/WriteError.html

Really thorough test of the index creation.

src/test/java/uk/nhs/hee/tis/trainee/forms/api/FormRPartAResourceTest.java
Judge40: Good spot, its ended up testing the wrong thing after my change and was passing because of POSTing with an ID set rather than passing because the Mongo exception was well handled (there's a case for monitoring branch coverage instead of line coverage in there 😁 ).

I've added the `null` ID back in and changed the verifications to reflect what we actually expect to happen.

Judge40:
> Misc thoughts:
> I wondered if the ExceptionHandler would be more maintainable or too broad if it looked an `errorCategory() == ErrorCategory.DUPLICATE_KEY` rather than `errorCode() == 11000` https://mongodb.github.io/mongo-java-driver/ {not sure of version} /javadoc/com/mongodb/WriteError.html
> 
> Really thorough test of the index creation.

Definitely a better solution, I looked for constants for the error codes and didn't find any. Didn't even spot the error category doing it for me 🤦🏻

JosephKelly (approved):
I'm trusting Sonar's 100% coverage that we're testing a Form's id is either set or not set for a PUT and POST, respectively

PR #279, opened 2021-11-18
Test/improve coverage

Judge40:


src/test/java/uk/nhs/hee/tis/trainee/forms/service/FormRPartBServiceImplTest.java
```diff
@@ -437,4 +537,25 @@ void shouldGetDraftFormRPartBById() {
...
+    output.setIsHealthy(input.getIsHealthy());
+    output.setHealthStatement(input.getHealthStatement());
+    output.setHavePreviousDeclarations(input.getHavePreviousDeclarations());
+    output.setPreviousDeclarations(input.getPreviousDeclarations());
+    output.setPreviousDeclarationSummary(input.getPreviousDeclarationSummary());
+    output.setHaveCurrentDeclarations(input.getHaveCurrentDeclarations());
+    output.setCurrentDeclarations(input.getCurrentDeclarations());
+    output.setCurrentDeclarationSummary(input.getCurrentDeclarationSummary());
+    return output;
+  }
```
JosephKelly, line 560: `BeanUtils.copyProperties()` here too.

src/test/java/uk/nhs/hee/tis/trainee/forms/service/FormRPartAServiceImplTest.java
```diff
@@ -242,4 +314,13 @@ void shouldGetDraftFormRPartAById() {
...
   }
+
+  private static FormRPartA copyForm(FormRPartA input) {
+    FormRPartA output = new FormRPartA();
+    output.setId(input.getId());
+    output.setTraineeTisId(input.getTraineeTisId());
+    output.setForename(input.getForename());
+    output.setSurname(input.getSurname());
+    return output;
+  }
```
JosephKelly, line 325: If I have to find something, you could reduce the size of the class be replacing this with a call to `org.springframework.beans.BeanUtils.copyProperties(input, new FormRPartA())`.

src/main/java/uk/nhs/hee/tis/trainee/forms/service/impl/FormRPartAServiceImpl.java
```diff
@@ -77,11 +77,10 @@ public FormRPartADto save(FormRPartADto formRPartADto) {
...
         || formRPartA.getLifecycleState() == LifecycleState.UNSUBMITTED) {
       cloudObjectRepository.save(formRPartA);
-      //Save in mongo for backward compatibility
-      repository.save(formRPartA);
-    } else {
-      formRPartA = repository.save(formRPartA);
     }
+
+    // Forms stored in cloud are still stored to Mongo for backwards compatibility.
+    formRPartA = repository.save(formRPartA);
```
JosephKelly, line 83: There is a subtle difference in returning the object saved in the `repository` rather than the parameter.  I think it's ok as I'm not aware of any further code that relies on the object identity.

I do prefer this simplification.

JosephKelly (commented):
Good spot on

src/main/java/uk/nhs/hee/tis/trainee/forms/service/impl/FormRPartAServiceImpl.java
Judge40: I changed the behaviour based on the tests, they were written to verify that the ID is set but were written in such a way that the ID was being set by the test code rather than by the implementation.

src/test/java/uk/nhs/hee/tis/trainee/forms/service/FormRPartAServiceImplTest.java
Judge40: Nice, didn't know of that util. I'll make the changes.

ReubenRobertsHEE (approved):
LGTM

PR #303, opened 2022-01-27
chore: delete non-ARCP pilot forms

Judge40:
Trainees who took part in the initial pilot created forms which were not
used for official ARCP submissions.
Delete all forms created before the second pilot (2021-09-01).

TIS21-2550


Judge40:
@Health-Education-England/full-stack I'm not sure whether we want/need a naming convention for Mongock migration files, that choice could be deferred though as, unlike Flyway, the file naming is not important.

Marcellofabbri (approved):

PR #323, opened 2022-03-23
fix: recalculate total leave on all FormR Part Bs

Judge40:
The `totalLeave` field was not correctly populated for some forms,
create a migrator to recalculate the field any record where it is zero.

TIS21-2791

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,7 +10,7 @@ plugins {
 }
 
 group = "uk.nhs.hee.tis.trainee"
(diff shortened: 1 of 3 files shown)


john12321 (approved):

Judge40:
Converted to draft while we consider how to update the S3 copies of the forms.

john12321 (approved):

ReubenRobertsHEE (approved):
LGTM

PR #346, opened 2022-05-17
chore(support): delete test FormR PartB

JosephKelly:
Form was submitted by dev team to validate user data.

NO-CARD: Support query tracked in spreadsheet


Judge40 (commented):
I think we should either include S3 deletion in this migration, or do it all off the books to keep some form of consistency in what we action.

ReubenRobertsHEE (dismissed):
LGTM

JosephKelly:
> I think we should either include S3 deletion in this migration, or do it all off the books to keep some form of consistency in what we action.

Drats... distraction meant I didn't mention I've removed from S3.

PR #351, opened 2022-05-25
fix: handle null work endDates in sort

ReubenRobertsHEE:
TIS21-2911: Sort work placements for previously-submitted forms

### DIFF
--- a/src/main/java/uk/nhs/hee/tis/trainee/forms/migration/SortWorkPlacements.java
+++ b/src/main/java/uk/nhs/hee/tis/trainee/forms/migration/SortWorkPlacements.java
@@ -60,7 +60,8 @@ public class SortWorkPlacements {
    */
   @Execution
   public void migrate() {
-    Comparator<Work> compareEndDates = Comparator.comparing(Work::getEndDate).reversed();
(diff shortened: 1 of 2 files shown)


src/main/java/uk/nhs/hee/tis/trainee/forms/migration/SortWorkPlacements.java
```diff
@@ -60,7 +60,8 @@ public SortWorkPlacements(MongoTemplate mongoTemplate, FormRPartBService formSer
    */
   @Execution
   public void migrate() {
-    Comparator<Work> compareEndDates = Comparator.comparing(Work::getEndDate).reversed();
+    Comparator<Work> compareEndDates = Comparator.comparing(Work::getEndDate,
+        Comparator.nullsLast(Comparator.naturalOrder())).reversed();
```
JosephKelly, line 64: This is equivalent to `Comparator.nullsFirst(Comparator.reverseOrder()))`.

Did you want to preserve/make clear that it is ordered in a particular way and then reversed or is the alternative a more accurate way of describing the intended ordering?

PR #352, opened 2022-05-27
chore: remove trainee form submitted in error

ReubenRobertsHEE:
This seems a rather painstaking way to do this, but it follows the
existing pattern of deleting e.g. the Test form.

NO-TICKET: https://hee-nhs-tis.slack.com/archives/G010E91L89K/p1653641991735899


ReubenRobertsHEE:
It's a little tricky figuring out which form was submitted first from the database records, but the S3 bucket makes it a bit easier: https://s3.console.aws.amazon.com/s3/buckets/tis-trainee-documents-upload-prod?prefix=160419/forms/formr-a/&region=eu-west-2

ReubenRobertsHEE:
Done manually

PR #447, opened 2023-01-03
Fix/standardise form ids

Judge40:
Marking as draft as I need to write some further tests and the application config needs the temp local hack removing from the database connection URI before merge.


Judge40:
The changes made here were split out in to several PRs, this is no longer needed.

PR #448, opened 2023-01-04
fix: create consistent form ids

Judge40:
The form ID number is a different format depending on whether the form The form ID number is a different format depending on whether the form was ever saved as a draft or not, direct submission results in a UUID style ID e.g. c95f2312-a08c-46d4-8e0c-5c13215c96d0 but if the form was saved as a draft before later submitting the ID will be in the ObjectId style e.g. 63a3376e7199d510f8f56808

Use an event listener to populate the form ID with a UUID string representation before converting it to a document, this avoids Mongo from setting the ObjectId automatically.
A follow up submission will migrate the existing ObjectIds to the new standardised UUID string format.

TIS21-4006

[TIS21-4006]: https://hee-tis.atlassian.net/browse/TIS21-4006?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,7 +10,7 @@ plugins {
 }
 
 group = "uk.nhs.hee.tis.trainee"
-version = "0.5.5"
+version = "0.5.6"
 
 configurations {
   compileOnly {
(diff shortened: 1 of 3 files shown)


ReubenRobertsHEE (approved):
Looks good

PR #449, opened 2023-01-04
feat: convert ObjectIds to UUID strings

Judge40:
All new forms are created with UUID strings as the ID, but legacy forms which went through the draft lifecycle state were created with an ObjectId instead of UUID.

Create a migrator to find all ObjectId based forms and convert them to use a UUID string based ID.

TIS21-4006
TIS21-4026

[TIS21-4006]: https://hee-tis.atlassian.net/browse/TIS21-4006?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TIS21-4026]: https://hee-tis.atlassian.net/browse/TIS21-4026?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,7 +10,7 @@ plugins {
 }
 
 group = "uk.nhs.hee.tis.trainee"
-version = "0.5.6"
+version = "0.6.0"
 
 configurations {
   compileOnly {
--- a/src/main/java/uk/nhs/hee/tis/trainee/forms/migration/ConvertObjectIdsToUuidStrings.java
+++ b/src/main/java/uk/nhs/hee/tis/trainee/forms/migration/ConvertObjectIdsToUuidStrings.java
@@ -104,16 +104,16 @@ public class ConvertObjectIdsToUuidStrings {
       form.setId(uuid);
       log.info("UUID {} was generated for form {}.", uuid, originalId);
 
-      // Save the updated form to the database and, if applicable, S3.
+      // Save the updated form to the database and S3.
       saveNewForm(form);
 
-      // Delete the existing form from the database and, if the ID format changed, S3.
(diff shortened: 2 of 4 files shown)


Judge40:
I've created this as a draft as I haven't had an opportunity to test it with a large data set to gauge performance, but it is ready to be reviewed while I'm doing that.

ReubenRobertsHEE:
This all looks pretty solid to me. If I had to change anything, it would be removing the 'if applicable' and 'if the ID format changed' from the various comments in the migrator, since the query will limit the processing to records which comply :nitpick: 😁

Judge40:
> This all looks pretty solid to me. If I had to change anything, it would be removing the 'if applicable' and 'if the ID format changed' from the various comments in the migrator, since the query will limit the processing to records which comply :nitpick: 😁

Bad copy paste from the original migrator, I thought I'd updated everything already. Will fix up those comments.

Judge40:
I've changed the logging level because it was generating a lot of noise in the logs during this migration, it doesn't feel like we should have permanent granular logging enabled anyway. If we regularly need more logged than we get with these new settings then we should instead look at changing how we log, rather than what levels we log by default.

Judge40:
Running the migrator locally took approximately 42 minutes running locally, I expect it to take significantly less time running in AWS but to be extra safe I've increased the health check grace period to 1 hour so the migrator won't be prematurely terminated.

ReubenRobertsHEE (approved):
👍

PR #451, opened 2023-01-05
refactor: make UUID migrator use generic document

Judge40:
The migrator was based on using AbstractForm type to update the form ID, however the follow up changes to make the ID field use UUID will be a breaking change to this migrator.
Refactor the migrator to use generic Documents instead, this will avoid having to fix the implementation later. The test's will still fail to compile and will need further modification during the next steps.

TIS21-4006
TIS21-4026

### DIFF
--- a/src/test/java/uk/nhs/hee/tis/trainee/forms/migration/ConvertObjectIdsToUuidStringsTest.java
+++ b/src/test/java/uk/nhs/hee/tis/trainee/forms/migration/ConvertObjectIdsToUuidStringsTest.java
@@ -45,6 +45,7 @@ import org.mockito.ArgumentCaptor;
 import org.mockito.internal.verification.Times;
 import org.springframework.core.env.Environment;
 import org.springframework.data.mongodb.core.MongoTemplate;
+import org.springframework.data.mongodb.core.convert.MappingMongoConverter;
 import org.springframework.data.mongodb.core.query.Query;
 import org.springframework.test.util.ReflectionTestUtils;
 import uk.nhs.hee.tis.trainee.forms.dto.FormRPartADto;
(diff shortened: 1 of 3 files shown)


src/test/java/uk/nhs/hee/tis/trainee/forms/migration/ConvertObjectIdsToUuidStringsTest.java
```diff
@@ -89,6 +93,18 @@ void setUp() {
...
     when(template.getCollectionName(FormRPartB.class)).thenReturn(PART_B_COLLECTION_NAME);
+
+    MappingMongoConverter converter = mock(MappingMongoConverter.class);
+    when(converter.read(any(), any())).thenAnswer(args -> {
+      Class<? extends AbstractForm> formClass = args.getArgument(0);
+      Document document = args.getArgument(1);
+
+      AbstractForm form = formClass.getConstructor().newInstance();
+      form.setId(document.get(ID_FIELD).toString());
+      form.setTraineeTisId(document.getString(TRAINEE_ID_FIELD));
```
ReubenRobertsHEE, line 104: Is it just the two lines above that will need to be rewritten later (I assume we're not changing the DTO id to UUID as well)?

ReubenRobertsHEE (approved):
Seems fair enough if we have to tweak the tests later; nice that the migrator itself will not need modification.

src/test/java/uk/nhs/hee/tis/trainee/forms/migration/ConvertObjectIdsToUuidStringsTest.java
Judge40: Yeah, just line 103.

I'm a bit undecided about the DTO, we can change it without any issues and it doesn't technically break the contract since it gets converted to a string in the JSON response.

If we change the DTO then the API call will fail instantly if the provided ID isn't UUID compatible, if we don't change the DTO then it would fail during the conversion from DTO to Entity instead... so it really comes down to where we want validation/failure to occur.

PR #452, opened 2023-01-06
feat: migrate AbstractForm from string to UUID

Judge40:
Migrate the AbstractForm's `id` field from String to UUID based. Create a Mongock migration to find all String based IDs and perform the conversion to UUIDs in the database.

Modify the entity to use UUID and update the required classes and tests.

TIS21-4006
TIS21-4027


ReubenRobertsHEE:
I see we are going to have lots of fun from Sonar about the casing of FormRPart[x] variables 🤣

ReubenRobertsHEE (approved):
Tested locally, worked for me

PR #455, opened 2023-01-13
feat: partial delete form

DorisWongNHS:
- listen to the delete event queue to handle partial delete in DB
- add `deleteType` and `fixedFields` in S3 Metadata when new forms are saved
- Mongock migration to bulk add `deleteType` and `fixedFields` to Metadata of all the existing forms

TIS21-4009
TIS21-4019
TIS21-4077
TIS21-4023

### DIFF
--- a/src/main/java/uk/nhs/hee/tis/trainee/forms/event/FormDeleteEventListener.java
+++ b/src/main/java/uk/nhs/hee/tis/trainee/forms/event/FormDeleteEventListener.java
@@ -46,7 +46,7 @@ public class FormDeleteEventListener {
   }
 
   /**
-   * Listener for receiving form delete event from SQS queue
+   * Listener for receiving form delete event from SQS queue.
    */
   @SqsListener("${application.aws.sqs.delete-event}")
   public void getFormDeleteEvent(String message) throws IOException {
--- a/src/main/java/uk/nhs/hee/tis/trainee/forms/event/FormDeleteEventListener.java
+++ b/src/main/java/uk/nhs/hee/tis/trainee/forms/event/FormDeleteEventListener.java
@@ -1,0 +1,86 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright 2023 Crown Copyright (Health Education England)
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of this software and
+ * associated documentation files (the "Software"), to deal in the Software without restriction,
+ * including without limitation the rights to use, copy, modify, merge, publish, distribute,
+ * sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all copies or
+ * substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT
+ * NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+ * DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package uk.nhs.hee.tis.trainee.forms.event;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.awspring.cloud.messaging.listener.annotation.SqsListener;
+import java.io.IOException;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.stereotype.Component;
+import uk.nhs.hee.tis.trainee.forms.dto.DeleteEventDto;
+import uk.nhs.hee.tis.trainee.forms.dto.enumeration.DeleteType;
+import uk.nhs.hee.tis.trainee.forms.service.FormRPartAService;
+import uk.nhs.hee.tis.trainee.forms.service.FormRPartBService;
+import uk.nhs.hee.tis.trainee.forms.service.exception.ApplicationException;
+
+@Slf4j
+@Component
+public class FormDeleteEventListener {
+
+  private final FormRPartAService formRPartAService;
+  private final FormRPartBService formRPartBService;
+
+  FormDeleteEventListener(FormRPartAService formRPartAService,
+      FormRPartBService formRPartBService) {
+    this.formRPartAService = formRPartAService;
+    this.formRPartBService = formRPartBService;
+  }
+
+  /**
+   * Listener for receiving form delete event from SQS queue
+   */
+  @SqsListener("${application.aws.sqs.delete-event}")
+  public void getFormDeleteEvent(String message) throws IOException {
+    try {
+      log.info("Form delete event received: {}", message);
+      ObjectMapper objectMapper = new ObjectMapper();
+      DeleteEventDto deleteEvent = objectMapper.readValue(message, DeleteEventDto.class);
+
+
+      if (deleteEvent.getDeleteType() == DeleteType.PARTIAL) {
+        final var eventDetails =
+            deleteEvent.getKey().split("/");
+        final var formId = eventDetails[3].split(".json")[0];
+        final var traineeTisId = eventDetails[0];
+        final var fixFields = deleteEvent.getFixedFields();
+
+        if (eventDetails[2].equals("formr-a")) {
+          formRPartAService.partialDeleteFormRPartAById(formId, traineeTisId, fixFields);
+          log.info("Partial delete successfully for trainee {} with form Id {} (FormRPartA)",
+              traineeTisId, formId);
+        } else if (eventDetails[2].equals("formr-b")) {
+          log.info("trainee: {}, Form B", eventDetails[0]);
+          formRPartBService.partialDeleteFormRPartBById(formId, traineeTisId, fixFields);
+          log.info("Partial delete successfully for trainee {} with form Id {} (FormRPartB)",
+              traineeTisId, formId);
+        }
+      } else {
+        log.error("Unexpected deleteType of form: {}" + deleteEvent.getDeleteType());
+        throw new ApplicationException("Unexpected deleteType of form: "
+            + deleteEvent.getDeleteType());
+      }
+    } catch (Exception e) {
+      log.error("Fail to delete form: {}", e);
+      throw new ApplicationException("Fail to delete form:", e);
+    }
+  }
+}
--- a/src/main/java/uk/nhs/hee/tis/trainee/forms/event/FormDeleteEventListener.java
+++ b/src/main/java/uk/nhs/hee/tis/trainee/forms/event/FormDeleteEventListener.java
@@ -1,5 +1,5 @@
 /*
- * The MIT License (MIT)
+ * The MIT License (MIT).
  *
  * Copyright 2023 Crown Copyright (Health Education England)
  *
@@ -32,6 +32,9 @@ import uk.nhs.hee.tis.trainee.forms.service.FormRPartAService;
 import uk.nhs.hee.tis.trainee.forms.service.FormRPartBService;
 import uk.nhs.hee.tis.trainee.forms.service.exception.ApplicationException;
 
+/**
+ * Listener for receiving form delete event from SQS queue.
+ */
 @Slf4j
 @Component
 public class FormDeleteEventListener {
@@ -46,10 +49,10 @@ public class FormDeleteEventListener {
   }
 
   /**
-   * Listener for receiving form delete event from SQS queue.
+   * Listener for handling form delete form event.
    */
   @SqsListener("${application.aws.sqs.delete-event}")
-  public void getFormDeleteEvent(String message) throws IOException {
+  public void handleFormDeleteEvent(String message) throws IOException {
     try {
       log.info("Form delete event received: {}", message);
       ObjectMapper objectMapper = new ObjectMapper();
@@ -68,7 +71,6 @@ public class FormDeleteEventListener {
           log.info("Partial delete successfully for trainee {} with form Id {} (FormRPartA)",
               traineeTisId, formId);
         } else if (eventDetails[2].equals("formr-b")) {
-          log.info("trainee: {}, Form B", eventDetails[0]);
           formRPartBService.partialDeleteFormRPartBById(formId, traineeTisId, fixFields);
           log.info("Partial delete successfully for trainee {} with form Id {} (FormRPartB)",
               traineeTisId, formId);
--- a/src/main/java/uk/nhs/hee/tis/trainee/forms/event/FormDeleteEventListener.java
+++ b/src/main/java/uk/nhs/hee/tis/trainee/forms/event/FormDeleteEventListener.java
@@ -24,6 +24,7 @@ package uk.nhs.hee.tis.trainee.forms.event;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.awspring.cloud.messaging.listener.annotation.SqsListener;
 import java.io.IOException;
+import java.util.Set;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Component;
 import uk.nhs.hee.tis.trainee.forms.dto.DeleteEventDto;
@@ -41,11 +42,14 @@ public class FormDeleteEventListener {
 
   private final FormRPartAService formRPartAService;
   private final FormRPartBService formRPartBService;
+  private final ObjectMapper objectMapper;
 
   FormDeleteEventListener(FormRPartAService formRPartAService,
-      FormRPartBService formRPartBService) {
+      FormRPartBService formRPartBService,
+      ObjectMapper objectMapper) {
     this.formRPartAService = formRPartAService;
     this.formRPartBService = formRPartBService;
+    this.objectMapper = objectMapper;
   }
 
   /**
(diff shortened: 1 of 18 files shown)


src/main/java/uk/nhs/hee/tis/trainee/forms/event/FormDeleteEventListener.java
```diff
@@ -0,0 +1,88 @@
...
+          log.info("Partial delete successfully for trainee {} with form Id {} (FormRPartA)",
+              traineeTisId, formId);
+        } else if (eventDetails[2].equals("formr-b")) {
+          formRPartBService.partialDeleteFormRPartBById(formId, traineeTisId, fixFields);
+          log.info("Partial delete successfully for trainee {} with form Id {} (FormRPartB)",
+              traineeTisId, formId);
+        }
+      } else {
+        log.error("Unexpected deleteType of form: {}" + deleteEvent.getDeleteType());
+        throw new ApplicationException("Unexpected deleteType of form: "
```
Judge40, line 80: Is the intention to send non-`PARTIAL` deletes to the DLQ for review?

src/main/java/uk/nhs/hee/tis/trainee/forms/service/impl/FormRPartAServiceImpl.java
```diff
@@ -110,4 +118,40 @@ public FormRPartADto getFormRPartAById(String id, String traineeTisId) {
...
+  public FormRPartADto partialDeleteFormRPartAById(
+      String id, String traineeTisId, String[] fixedFields) {
+    log.info("Request to partial delete FormRPartA by id : {}", id);
+
+    try {
+      FormRPartA formRPartA = repository.findByIdAndTraineeTisId(UUID.fromString(id), traineeTisId)
+          .orElse(null);
+
+      if (formRPartA != null) {
+        ObjectMapper objectMapper = new ObjectMapper().findAndRegisterModules();
```
Judge40, line 134: This should probably be injected via the constructor.

src/main/java/uk/nhs/hee/tis/trainee/forms/service/impl/FormRPartAServiceImpl.java
```diff
@@ -110,4 +118,40 @@ public FormRPartADto getFormRPartAById(String id, String traineeTisId) {
...
+        JsonNode jsonForm = objectMapper.valueToTree(formRPartA);
+
+        for (Iterator<String> fieldIterator = jsonForm.fieldNames(); fieldIterator.hasNext(); ) {
+          String fieldName = fieldIterator.next();
+
+          if (!Set.of(fixedFields).contains(fieldName)) {
```
Judge40, line 140: `Set.of(fixedFields)` here will create a new `Set` on every loop iteration, I recommend initializing the set outside of the loop. Alternatively, changing the method parameter to require a Set might be clearer.

src/main/java/uk/nhs/hee/tis/trainee/forms/service/impl/FormRPartBServiceImpl.java
```diff
@@ -105,4 +113,42 @@ public FormRPartBDto getFormRPartBById(String id, String traineeTisId) {
...
+          .orElse(null);
+
+      if (formRPartB != null) {
+        ObjectMapper objectMapper = new ObjectMapper().findAndRegisterModules();
+        JsonNode jsonForm = objectMapper.valueToTree(formRPartB);
+
+        for (Iterator<String> fieldIterator = jsonForm.fieldNames(); fieldIterator.hasNext(); ) {
+          String fieldName = fieldIterator.next();
+
+          if (!Set.of(fixedFields).contains(fieldName)) {
```
Judge40, line 137: Same as for PartA

src/main/java/uk/nhs/hee/tis/trainee/forms/service/impl/FormRPartBServiceImpl.java
```diff
@@ -105,4 +113,42 @@ public FormRPartBDto getFormRPartBById(String id, String traineeTisId) {
...
+      String id, String traineeTisId, String[] fixedFields) {
+    log.info("Request to partial delete FormRPartB by id : {}", id);
+
+    try {
+      FormRPartB formRPartB = formRPartBRepository
+          .findByIdAndTraineeTisId(UUID.fromString(id), traineeTisId)
+          .orElse(null);
+
+      if (formRPartB != null) {
+        ObjectMapper objectMapper = new ObjectMapper().findAndRegisterModules();
```
Judge40, line 131: Same as for PartA

src/test/java/uk/nhs/hee/tis/trainee/forms/event/FormDeleteEventListenerTest.java
```diff
@@ -0,0 +1,116 @@
...
+import java.io.IOException;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.junit.jupiter.MockitoExtension;
+import uk.nhs.hee.tis.trainee.forms.service.FormRPartAService;
+import uk.nhs.hee.tis.trainee.forms.service.FormRPartBService;
+import uk.nhs.hee.tis.trainee.forms.service.exception.ApplicationException;
+
+@ExtendWith(MockitoExtension.class)
```
Judge40, line 40: We're not using any mockito annotations that I could see, see this extension is not required.

src/main/java/uk/nhs/hee/tis/trainee/forms/event/FormDeleteEventListener.java
```diff
@@ -0,0 +1,88 @@
...
+
+        if (eventDetails[2].equals("formr-a")) {
+          formRPartAService.partialDeleteFormRPartAById(formId, traineeTisId, fixFields);
+          log.info("Partial delete successfully for trainee {} with form Id {} (FormRPartA)",
+              traineeTisId, formId);
+        } else if (eventDetails[2].equals("formr-b")) {
+          formRPartBService.partialDeleteFormRPartBById(formId, traineeTisId, fixFields);
+          log.info("Partial delete successfully for trainee {} with form Id {} (FormRPartB)",
+              traineeTisId, formId);
+        }
```
Judge40, line 77: We should log if we get an unknown form type too.

Play around and see if a switch case is more readable or not
```java
switch (eventDetails[2]) {
  case "formr-a":
    // Do part a stuff
  case "formr-b":
    // Do part b stuff
  default:
    log.info("unknown form type");
}
````

File: src/main/java/uk/nhs/hee/tis/trainee/forms/event/FormDeleteEventListener.java [hunk 8]
```diff
@@ -0,0 +1,88 @@
...
+  /**
+   * Listener for handling form delete form event.
+   */
+  @SqsListener("${application.aws.sqs.delete-event}")
+  public void handleFormDeleteEvent(String message) throws IOException {
+    try {
+      log.info("Form delete event received: {}", message);
+      ObjectMapper objectMapper = new ObjectMapper();
+      DeleteEventDto deleteEvent = objectMapper.readValue(message, DeleteEventDto.class);
+
```
Judge40, line 60: Formatting: double line break

src/main/java/uk/nhs/hee/tis/trainee/forms/event/FormDeleteEventListener.java
```diff
@@ -0,0 +1,88 @@
...
+
+      if (deleteEvent.getDeleteType() == DeleteType.PARTIAL) {
+        final var eventDetails =
```
Judge40, line 63: Quite a lot of `final var`/`var` usage here that is making it hard to follow outside of an IDE, consider switching to standard type for the declarations unless its something with ridiculous generics all over the place.

src/main/java/uk/nhs/hee/tis/trainee/forms/migration/AddPartialDeleteMetadataToS3.java
```diff
@@ -0,0 +1,101 @@
...
+  /**
+   * Find all object on S3 and add metadata.
+   */
+  @Execution
+  public void migrate() {
+    log.info("Starting migration to add partial delete related metadata to existing forms on S3.");
+
+    try {
+      List<S3ObjectSummary> keyList = new ArrayList<>();
+      var objects = amazonS3.listObjects(bucketName);
```
Judge40, line 66: As above, the heavy use of `var` makes this harder to follow than it should be. There is only a single variable with a clear type, I don't know what any of the AWS SDK calls are returning.

src/test/java/uk/nhs/hee/tis/trainee/forms/event/FormDeleteEventListenerTest.java
```diff
@@ -0,0 +1,116 @@
...
+  @BeforeEach
+  void setUp() {
+    formRPartAService = mock(FormRPartAService.class);
+    formRPartBService = mock(FormRPartBService.class);
+    listener = new FormDeleteEventListener(formRPartAService, formRPartBService);
+  }
+
+  @Test
+  void shouldPartialDeleteFormA() throws IOException {
+    final String MESSAGE = "{\"deleteType\":\"PARTIAL\","
```
Judge40, line 56: Take advantage of text blocks to make this more readable

e.g.
```java
String message = """
  {
    "deleteType": "PARTIAL",
    "bucket": "document-upload",
    "key": "1/forms/formr-a/1000a.json",
    "fixedFields": ["id", "traineeTisId"]
  }
""";
```

src/main/java/uk/nhs/hee/tis/trainee/forms/migration/AddPartialDeleteMetadataToS3.java
```diff
@@ -0,0 +1,101 @@
...
+      for (var obj : keyList) {
+        final var object = amazonS3.getObject(bucketName, obj.getKey());
+
+        var metadata = object.getObjectMetadata();
+        metadata.addUserMetadata("deletetype", DeleteType.PARTIAL.name());
+        metadata.addUserMetadata("fixedfields", FIXED_FIELDS);
+
+        final var request = new PutObjectRequest(
+            bucketName, obj.getKey(), object.getObjectContent(), metadata);
+        amazonS3.putObject(request);
```
Judge40, line 84: To make this more rerunnable (on failure) you should add a check here for whether the `deleteType`/`fixedFields` metadata already exists. If it does then we skip updating the object.

src/main/java/uk/nhs/hee/tis/trainee/forms/migration/AddPartialDeleteMetadataToS3.java
```diff
@@ -0,0 +1,101 @@
...
+      }
+
+      log.info("Finish migration");
+    }
+    catch (Exception e) {
+      log.error("Fail to add metadata to existing forms in bucket " + bucketName + ": ", e);
```
Judge40, line 90: Given there zero downtime implications for this migrator, I'd maybe let the exception be thrown to stop the migrator.

Already migrated forms should not be reverted, but it gives us a clear indication that we need to fix and re-run rather than having to manually fix afterwards with no option to rerun the migration (without DB fiddling). Combined with the metadata check above we should have a fairly streamlined fail and retry process.

src/test/java/uk/nhs/hee/tis/trainee/forms/migration/AddPartialDeleteMetadataToS3Test.java
```diff
@@ -0,0 +1,195 @@
...
+    when(s3.getObject(BUCKET_NAME, "3")).thenReturn(object3);
+
+    migration.migrate();
+
+    verify(s3, times(3)).putObject(putRequestCaptor.capture());
+    List<PutObjectRequest> putObjectRequests = putRequestCaptor.getAllValues();
+    assertThat("Unexpected put object request count.",
+        putObjectRequests.size(), CoreMatchers.is(3));
+
+    var assertId = 1;
```
Judge40, line 152: This is an unnecessary use of `var` vs `int`

src/test/java/uk/nhs/hee/tis/trainee/forms/migration/AddPartialDeleteMetadataToS3Test.java
```diff
@@ -0,0 +1,195 @@
...
+          putObjectRequest.getBucketName(), is(BUCKET_NAME));
+      assertThat("Unexpected key.",
+          putObjectRequest.getKey(), is(Integer.toString(assertId)));
+      final var resultInputStream = putObjectRequest.getInputStream();
+      ObjectMapper mapper = new ObjectMapper();
+      Map<String, Object> resultJsonMap = mapper.readValue(resultInputStream, Map.class);
+      assertThat("Unexpected input stream.",
+          resultJsonMap.get("id"), is(Integer.toString(assertId)));
+
+      final var resultUserMetadata =
```
Judge40, line 164: As with other `var` comments, it's hard to know what this call returns outside of an IDE

src/test/java/uk/nhs/hee/tis/trainee/forms/service/FormRPartAServiceImplTest.java
```diff
@@ -321,4 +332,41 @@ void shouldGetDraftFormRPartAById() {
...
+        DEFAULT_ID_STRING, DEFAULT_TRAINEE_TIS_ID, FIXED_FIELDS);
+
+    verify(repositoryMock).save(formRPartACaptor.capture());
+    FormRPartA dto = formRPartACaptor.getValue();
+    assertThat("Unexpected form ID.", dto.getId(), is(DEFAULT_ID));
+    assertThat("Unexpected trainee ID.", dto.getTraineeTisId(), is(DEFAULT_TRAINEE_TIS_ID));
+    assertThat("Unexpected lifecycle states.",
+        dto.getLifecycleState(), is(LifecycleState.DELETED));
+    assertThat("Unexpected forename.", dto.getForename(), is(nullValue()));
+    assertThat("Unexpected surname.", dto.getSurname(), is(nullValue()));
```
Judge40, line 351: These should just be `nullValue()` instead of `is(nullValue())`.

src/test/java/uk/nhs/hee/tis/trainee/forms/service/FormRPartAServiceImplTest.java
```diff
@@ -321,4 +332,41 @@ void shouldGetDraftFormRPartAById() {
...
+  }
```
Judge40, line 352: I generally prefer individual field checks (like you've done) over building an "expectedResult" instance of the object being tested, but in this case it might be an appropriate way to test all other fields get nulled without having to test each and every field.

e.g.
```java
FormRPartA expectedDto = new FormRPartA();
expectedDto.setId(DEFAULT_ID);
expectedDto.setTraineeTisId(DEFAULT_TRAINEE_TIS_ID);
expectedDto.setLifeCycle(LifecycleState.DELETED);

assertThat("Unexpected DTO.", dto, is(expectedDto));
```

src/test/java/uk/nhs/hee/tis/trainee/forms/service/FormRPartAServiceImplTest.java
```diff
@@ -321,4 +332,41 @@ void shouldGetDraftFormRPartAById() {
...
+
+  @Test
+  void shouldNotPartialDeleteWhenFormRPartANotFoundInDb() {
+    when(repositoryMock.findByIdAndTraineeTisId(DEFAULT_ID, DEFAULT_TRAINEE_TIS_ID))
+        .thenReturn(Optional.ofNullable(null));
```
Judge40, line 357: Change to return `Optional.empty()` as a cleaner alternative.

src/test/java/uk/nhs/hee/tis/trainee/forms/service/FormRPartAServiceImplTest.java
```diff
@@ -321,4 +332,41 @@ void shouldGetDraftFormRPartAById() {
...
+
+    service.partialDeleteFormRPartAById(
+        DEFAULT_ID_STRING, DEFAULT_TRAINEE_TIS_ID, FIXED_FIELDS);
+
+    verify(repositoryMock, never()).save(formRPartACaptor.capture());
+  }
+
+  @Test
+  void shouldThrowExceptionWhenFailToPartialDeleteFormRPartA() throws ApplicationException {
+    doThrow(IllegalArgumentException.class)
```
Judge40, line 367: Generally we only need to use this technique for `void` methods, you should be able to still with a more typical
```
when(repositoryMock.findIdAndTraineeTisId(any()), any())).thenThrow(IllegalArgumentException.class);
```

src/test/java/uk/nhs/hee/tis/trainee/forms/service/FormRPartBServiceImplTest.java
```diff
@@ -587,4 +594,65 @@ void shouldGetDraftFormRPartBById() {
         dto.getHavePreviousUnresolvedDeclarations(),
         is(DEFAULT_HAVE_PREVIOUS_UNRESOLVED_DECLARATIONS));
   }
+
+  @Test
```
Judge40, line 598: Same comments as the Part A tests

Judge40 (changes requested):

src/main/java/uk/nhs/hee/tis/trainee/forms/event/FormDeleteEventListener.java
DorisWongNHS: Yeah, it is just for the edge cases though, trainee forms NON-PARTIAL deletion should not happen (in our current design) after mongock migration, it may be metadata problems / programme bugs that we want to look into. It would be good if we get alerts from DLQ.

DorisWongNHS:
🙈 Oh.. I found that I have misunderstood Sonarcloud, and thought it suggested me to use `var` after Java10. (lol)
https://rules.sonarsource.com/java/RSPEC-6212
Changed it back now.

src/main/java/uk/nhs/hee/tis/trainee/forms/migration/AddPartialDeleteMetadataToS3.java
DorisWongNHS: Good shout. The `addUserMetadata()` would do an "upsert" to update if metadata exists/ insert if not exists. It should keep a correct result even if the migration is rerun though, it's good to skip the unneeded putObject operation if it's already updated.

Judge40 (approved):

PR #521, opened 2023-07-05
feat: add arcpYear to FormRPartB

EdwardBarclay:


src/main/java/uk/nhs/hee/tis/trainee/forms/dto/FormRPartBDto.java
```diff
@@ -72,4 +72,5 @@ public class FormRPartBDto {
   private CovidDeclarationDto covidDeclarationDto;
   private Boolean haveCurrentUnresolvedDeclarations;
   private Boolean havePreviousUnresolvedDeclarations;
+  private Number arcpYear;
```
Judge40, line 75: I don't think I've ever seen the `Number` abstract actually used 😅 
Maybe use [Year](https://docs.oracle.com/javase/8/docs/api/java/time/Year.html) instead?

Judge40 (changes requested):
While this will get flagged as full coverage still, we need to check that the field is mapped properly between JSON and DTO (`FormRPartBResourceTest`) and DTO and Entity (both ways)(`FormRPartBServiceTest`)

There should be some existing tests you can add assertions to so you can make sure the mapper etc. is all working.

PR #538, opened 2023-08-08
feat: allow deletion of draft forms

Judge40:
In order to support the discarding of a draft form, the user must be allowed to delete any draft forms that they have created. Create a new DELETE endpoint with supporting service methods for both Part-A and Part-B forms.

TIS21-4922
TIS21-4877

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,7 +10,7 @@ plugins {
 }
 
 group = "uk.nhs.hee.tis.trainee"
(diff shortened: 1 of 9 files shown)


ReubenRobertsHEE:
Its always nice to see a code-formatting war :trollface:  😂

EdwardBarclay (approved):
LGTM

PR #585, opened 2023-12-20
feat: move form in DB and S3

DorisWongNHS:
- After confirm with Naz, hard delete will be perform on the old formR record when the form is moved.
- existing methods in tis-trainee-forms intended to s handle FormR Part A & B separately, mixing with specific logics. So I try to write separate code for move form at the end
- return bad request status to the endpoint if any exceptions occur, with error description in the header. Will try to return error message in Slack with Lambda in later step

TIS21-5437
TIS21-5506


build.gradle
```diff
@@ -10,7 +10,7 @@ plugins {
 }
 
 group = "uk.nhs.hee.tis.trainee"
-version = "0.12.0"
+version = "0.11.0"
```
ReubenRobertsHEE, line 13: typo I think :)

src/main/java/uk/nhs/hee/tis/trainee/forms/api/FormRelocateResource.java
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  /**
+   * POST  /form-relocate : Relocate a form (attach a form from one trainee to another).
+   *
+   * @param formId  The ID of the form that need to relocate
+   * @param targetTrainee  The TraineeTisId of the target trainee
+   * @return the status of relocation
+   */
+  @PostMapping("/form-relocate/{formId}")
```
ReubenRobertsHEE, line 56: I wonder whether a PATCH mapping might be more appropriate here (assuming that can be integrated with slack, of course!)

src/main/java/uk/nhs/hee/tis/trainee/forms/service/FormRelocateService.java
```diff
@@ -0,0 +1,186 @@
...
+      String sourceTrainee = formR.getTraineeTisId();
+      try {
+        updateTargetTraineeInDb(formR, targetTrainee);
+        if (formR.getLifecycleState() != LifecycleState.DRAFT) {
+          relocateFormInS3(formType, formId, sourceTrainee, targetTrainee);
+        }
+      } catch (Exception e) {
+        log.error("Fail to relocate FormR to target trainee: " + e + ". Rolling back...");
+        try {
+          updateTargetTraineeInDb(formR, sourceTrainee);
```
ReubenRobertsHEE, line 104: It feels like we should be able to use a transaction for this: I don't have any code to offer as a template, but it might be worth looking at the mongo ClientSession to see if that offers a neater way of handling this.

src/main/java/uk/nhs/hee/tis/trainee/forms/service/FormRelocateService.java
```diff
@@ -0,0 +1,186 @@
...
+      final PutObjectRequest request = new PutObjectRequest(
+          bucketName,
+          targetKey,
+          new ByteArrayInputStream(objectMapper.writeValueAsBytes(formr)),
+          metadata);
+      amazonS3.putObject(request);
+      log.info("Form R in S3 relocated from " + sourceKey + " to " + targetKey);
+
+      // Delete Form R from source trainee
+      amazonS3.deleteObject(bucketName, sourceKey);
```
ReubenRobertsHEE, line 182: On the ticket it says 'The old form is removed from the NDW (existing partial delete to keep a history is fine)' - this code looks like a real (not partial) delete. Is that right? I think it might need to be changed if so.

src/main/java/uk/nhs/hee/tis/trainee/forms/service/FormRelocateService.java
ReubenRobertsHEE: Actually, the more I look at this, the more it seems like it is actually fine: spanning a transaction over a DB and an AWS action might be a bit of a stretch!

src/main/java/uk/nhs/hee/tis/trainee/forms/service/FormRelocateService.java
DorisWongNHS: Thank you @ReubenRobertsHEE for the comment and discussion. 🌮 
It's a good shout. As you have mentioned, the ticket is promoting partial delete to keep history, while earlier conversation in the channel suggested deleting the forms directly from S3 to avoid NDW clashes: [chat link](https://hee-nhs-tis.slack.com/archives/G0135LS4JVA/p1698248857703269?thread_ts=1698246122.129819&cid=G0135LS4JVA)
@Judge40 do you have it in mind whether we should partial or hard delete this in S3? If partial delete is conducted, does it mean we should halt NDW export and remove the messages from the SQS queue in Lambda to avoid clash?

src/main/java/uk/nhs/hee/tis/trainee/forms/service/FormRelocateService.java
DorisWongNHS: I was struggling to implement roll back in a tidy way as it needs to perform across DB and S3 operations. this is the version I have tried my best to make it simple.
I tried to use `@transactional` annotation, but it doesn't seem to work on these internal method calls within the same class. Any ideas about rolling back are welcomed 👂

src/main/java/uk/nhs/hee/tis/trainee/forms/service/FormRelocateService.java
Judge40: It's a question for Naz to answer really, since it affects what users will see (or not see) on downstream reports.

Depending on the decision:

 - Hard delete: no issue here since deletes don't send an S3 notification to the exporter. Possibly risk if that ever changed.
 - Soft delete: generating a new form ID seems like the sensible option to avoid race conditions and/or messing with queues. Having multiple forms with the same uuid is bound to cause us an issue at some point and just feels wrong.

Another thing we might want to consider is whether we should be setting a "delete reason" on soft deleted forms. In our scenarios, that would be with a value of either
1. DELETED
2. MOVED

Do we also need a record of where a form was moved to, or can we rely on logs if we ever need to answer that?

DorisWongNHS: As confirmed by Naz, we will hard delete the old formR record when the form is moved.

DorisWongNHS:
One of the new issues (code smells) in SonarCloud is about incorret order of method argument: [link](https://sonarcloud.io/project/issues?pullRequest=585&open=AYyHQnj0Np1srIiJi38d&id=Health-Education-England_tis-trainee-forms). I do that intentionally to input the sourceTrainee as a targetTrainee for the Rollback process. I would like to keep it although Sonar thinks it is a bug.

All other code smells are around the names with more than 1 consecutive capital letters. It happens to existing name like `FormRPartBRepository` and `FormRPartA` which is currently widely used in the whole service. I will keep it as is for this PR.

src/main/java/uk/nhs/hee/tis/trainee/forms/service/FormRelocateService.java
```diff
@@ -0,0 +1,182 @@
...
+      }
+    }
+  }
+
+  private AbstractForm getMoveFormInfoInDb(String formId) {
+    try {
+      Optional<FormRPartA> optionalFormRPartA =
+          formRPartARepository.findById(UUID.fromString(formId));
+      Optional<FormRPartB> optionalFormRPartB =
+          formRPartBRepository.findById(UUID.fromString(formId));
```
Judge40, line 119: Only call this if `optionalFormRPartA` is empty, otherwise it's just a wasted DB request.

src/main/java/uk/nhs/hee/tis/trainee/forms/service/FormRelocateService.java
```diff
@@ -0,0 +1,182 @@
...
+    formr.put("traineeTisId", targetTrainee);
+
+    // Save the form under target trainee
+    final PutObjectRequest request = new PutObjectRequest(
+        bucketName,
+        targetKey,
+        new ByteArrayInputStream(mapper.writeValueAsBytes(formr)),
+        metadata);
+    amazonS3.putObject(request);
+    log.info("Form R in S3 relocated from " + sourceKey + " to " + targetKey);
```
Judge40, line 176: Can the save to DB and S3 not re-use the existing `AbstractCloudRepository.save(AbstractForm)`?
I can't see a benefit to pulling the S3 object and having to modify the JSON vs just saving the Abstract form to JSON directly.

src/main/java/uk/nhs/hee/tis/trainee/forms/service/FormRelocateService.java
```diff
@@ -0,0 +1,182 @@
...
+      Optional<FormRPartB> optionalFormRPartB =
+          formRPartBRepository.findById(UUID.fromString(formId));
+
+      if (optionalFormRPartA.isPresent()) {
+        return optionalFormRPartA.get();
+      }
+      else if (optionalFormRPartB.isPresent()) {
+        return optionalFormRPartB.get();
+      }
+      return null;
```
Judge40, line 127: I'd prefer this method to return `Optional<AbstractForm>` instead of `null`

src/main/java/uk/nhs/hee/tis/trainee/forms/service/FormRelocateService.java
```diff
@@ -0,0 +1,182 @@
...
+        if (formR.getLifecycleState() != LifecycleState.DRAFT) {
+          relocateFormInS3(formType, formId, sourceTrainee, targetTrainee);
+        }
+      } catch (Exception e) {
+        log.error("Fail to relocate FormR to target trainee: " + e + ". Rolling back...");
+        try {
+          updateTargetTraineeInDb(formR, sourceTrainee);
+          if (formR.getLifecycleState() != LifecycleState.DRAFT) {
+            relocateFormInS3(formType, formId, targetTrainee, sourceTrainee);
+          }
```
Judge40, line 105: While only a few lines, extracting this in to a private method will make it easier to understand the code. I had to go back and read again to be clear on which variables were being switched around.

Something like this?
```java
try {
  performRelocate(formR, sourceTrainee, targetTrainee);
} catch (Exception e) {
  log.error("blah blah");
  
  try {
    performRelocate(formR, targetTrainee, sourceTrainee);
  } catch (Exception e) {
    log.error("blah blah");
  }
  throw new ExceptionBlahBlah();
}

...

private void performRelocate(AbstractForm form, String fromTraineeId, String toTraineeId) {
  updateTargetTraineeInDb(form, toTraineeId);
  
  if (form.getLifecycleState() != LifecycleState.DRAFT) {
    String formType = formR.getFormType();
    relocateFormInS3(formType, formId, fromTraineeId, toTraineeId);
  }
}

File: src/main/java/uk/nhs/hee/tis/trainee/forms/service/FormRelocateService.java
Comment (Judge40) [inline src/main/java/uk/nhs/hee/tis/trainee/forms/service/FormRelocateService.java:1]: Not particularly important, but the naming of the methods could be a little more generic to reduce refactoring when we need to include the relocation of other forms we implement.

File: src/main/java/uk/nhs/hee/tis/trainee/forms/api/FormRelocateResource.java [hunk 9]
```diff
@@ -0,0 +1,79 @@
...
+
+  /**
+   * PATCH  /form-relocate : Relocate a form (attach a form from one trainee to another).
+   *
+   * @param formId  The ID of the form that need to relocate
+   * @param targetTrainee  The TraineeTisId of the target trainee
+   * @return the status of relocation
+   */
+  @PatchMapping("/form-relocate/{formId}")
+  public ResponseEntity<Void> relocateFormR(@PathVariable String formId,
```
Judge40, line 57: Change the method name and some of the logs to be more generic, since this endpoint will be useful for non-FormR types in the future.

Judge40 (changes requested):

src/main/java/uk/nhs/hee/tis/trainee/forms/service/FormRelocateService.java
DorisWongNHS: >     try {
>       Optional<FormRPartA> optionalFormRPartA =
>           formRPartARepository.findById(UUID.fromString(formId));
>       if (optionalFormRPartA.isPresent()) {
>         return optionalFormRPartA.get();
>       }
>       else {
>         Optional<FormRPartB> optionalFormRPartB =
>             formRPartBRepository.findById(UUID.fromString(formId));
>         return optionalFormRPartB.get();
>       }
>     } catch (Exception e) {
>       log.error("Fail to get form with ID " + formId + ": " + e);
>       throw new ApplicationException("Fail to get form with ID " + formId + ": " + e.toString());
>     }

I was trying to return `Optional<AbstractForm>` but having issuing converting `Optional<FormRPartA>` into `Optional<AbstractForm>`. 
After refactoring, it is now returning AbstractForm with value optionalFormRPartA.get() or optionalFormRPartB.get(). When it is not found in both FormR Part A & B, it would intentionally throw exception. Please advise if there are better way to refactor it, please.

DorisWongNHS: Thank @Judge40. It is now updated to return `Optional<AbstractForm>` and handle if the option abstractForm is not present.

Judge40 (approved):

PR #586, opened 2023-12-21
Disable covid declaration feature flag

EdwardBarclay:
TIS21-5401

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -10,7 +10,7 @@ plugins {
 }
 
 group = "uk.nhs.hee.tis.trainee"
-version = "0.12.1"
+version = "0.13.0"
 
 configurations {
   compileOnly {
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -10,7 +10,7 @@ plugins {
 }
 
 group = "uk.nhs.hee.tis.trainee"
-version = "0.12.0"
+version = "0.12.1"
 
 configurations {
   compileOnly {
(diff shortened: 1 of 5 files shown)


src/main/java/uk/nhs/hee/tis/trainee/forms/api/FeatureFlagResource.java
```diff
@@ -49,6 +49,15 @@ public class FeatureFlagResource {
...
     log.debug("Get all the feature flags for forms");
+
+    // Create a new instance of FormRPartB and set covidDeclaration to false
+    FeatureConfigurationProperties.FormRPartB formRPartB =
+        new FeatureConfigurationProperties.FormRPartB();
+    formRPartB.setCovidDeclaration(false);
+
+    // Update the featureConfig object with the new FormRPartB instance
+    featureConfig.setFormRPartB(formRPartB);
+
```
Judge40, line 60: You shouldn't be modifying the config like this. The value is set in the application properties https://github.com/Health-Education-England/tis-trainee-forms/blob/main/src/main/resources/application.yml#L12

We could just set the env var to override it, but the default should probably be changed to false instead in this case.

build.gradle.kts
```diff
@@ -10,7 +10,7 @@ plugins {
 }
 
 group = "uk.nhs.hee.tis.trainee"
-version = "0.12.0"
+version = "0.12.1"
```
Judge40, line 13: This should be a minor version bump, since it's a functional change and not a bug fix.

Judge40 (changes requested):

Judge40 (approved):

PR #589, opened 2023-12-27
feat: temp re-enable covid declaration flag

EdwardBarclay:
until the logic for the FE around unsubmitted/draft forms is agreed

TIS21-5401

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,21 +34,20 @@ Error and exception logging is done using Sentry.
 
 #### Environmental Variables
 
-| Name                                  | Description                                                                    | Default   |
-|---------------------------------------|--------------------------------------------------------------------------------|-----------|
-| DB_HOST                               | The MongoDB host to connect to.                                                | localhost |
-| DB_PORT                               | The port to connect to MongoDB on.                                             | 27017     |
-| DB_USER                               | The username to access the MongoDB instance.                                   | admin     |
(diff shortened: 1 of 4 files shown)


DorisWongNHS:
The description and default value of the `FEATURE_FORMR_PARTB_COVID_DECLARATION` in README may need to be updated as well. https://github.com/Health-Education-England/tis-trainee-forms/blob/main/README.md?plain=1#L46

I am a bit confuse, other than the test, is this variable be used in any other place?

EdwardBarclay:
The variable is used in the front end to choose wether on not the covid declaration section is visiable when filling out a formR PartB form. it is sent through when forms are fetched as a feature flag variable

The value and description in the read-me can remain the same, it should have been changed in the previous PR but since its being reverted it can remain as is.

DorisWongNHS:
I see. 🌮  Thank you for the clarification about the usage. And I found how it works with `FeatureConfigurationProperties` from the historical commits.

Thank you for making sure the default value of `FEATURE_FORMR_PARTB_COVID_DECLARATION` in README is up-to-date. I was just wondering the variable description seems to be mixed up with the  `APPLICATION_FILESTORE_ALWAYSSTORE`.  I know it is an historical issue.

DorisWongNHS (approved):

DorisWongNHS (approved):

PR #590, opened 2023-12-29
feat: disable covid declaration flag

EdwardBarclay:
this will need to be merged at the same time as the FE PR (https://github.com/Health-Education-England/tis-trainee-ui/pull/995)

TIS21-5401

### DIFF
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -9,7 +9,7 @@ application:
 
 features:
   formr-partb:
-    covid-declaration: ${FEATURE_FORMR_PARTB_COVID_DECLARATION:true}
+    covid-declaration: ${FEATURE_FORMR_PARTB_COVID_DECLARATION:false}
 
 logging:
   level:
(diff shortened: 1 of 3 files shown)


src/main/resources/application.yml
```diff
@@ -9,7 +9,7 @@ application:
 
 features:
   formr-partb:
-    covid-declaration: ${FEATURE_FORMR_PARTB_COVID_DECLARATION:true}
+    covid-declaration: ${FEATURE_FORMR_PARTB_COVID_DECLARATION:false}
```
ReubenRobertsHEE, line 12: Is there a reason we don't just configure the environmental var FEATURE_FORMR_PARTB_COVID_DECLARATION to be false? It looks like it's not retrieved from the parameter store at the moment, but then I'm not sure why we have it as a configurable variable, if there is no way to give it a value other than the default?

(The default value is also mentioned in the readme.md, btw :))

Judge40: I'd argue that the env var was put in to be a "standard," implementation, but realistically its not something we would toggle on/off enough to warrant it really being an env var.

ReubenRobertsHEE (approved):

PR #594, opened 2024-01-11
Feat/form submitted field validation

ReubenRobertsHEE:
This is possibly a bit of a verbose solution, so feel free to request a rethink of the architecture.

There are quite a few (quite repetitive) tests, but generally they are focused on highlighting failed validation, not successful validation. There are two annotation validation classes for which there are explicit tests, since they have non-trivial logic.

The patched-in 'WTE cannot be null' rule from the PR earlier this week has been left in place, but in theory it should be removable at this point.

TIS21-5507
TIS21-5574
TIS21-5575


ReubenRobertsHEE:
Current error message from validation looks like:
`javax.validation.ConstraintViolationException: validateFormRPartA.formRPartADto.email: must be a well-formed email address, validateFormRPartA.formRPartADto.dateOfBirth: You must be at least 18 years old, validateFormRPartA.formRPartADto.forename: size must be between 1 and 100, validateFormRPartA.formRPartADto.telephoneNumber: Invalid telephone number, validateFormRPartA.formRPartADto.mobileNumber: Invalid telephone number, validateFormRPartA.formRPartADto.isLeadingToCct: must not be null`

ReubenRobertsHEE:
For the front-end team (ahem), it would be great if you could just give the revised DTOs a once-over to confirm the validation annotations match up with the existing front-end validation rules.

john12321:
> For the front-end team (ahem), it would be great if you could just give the revised DTOs a once-over to confirm the validation annotations match up with the existing front-end validation rules.

On behalf of the front-end team, I say no problem.

src/main/java/uk/nhs/hee/tis/trainee/forms/dto/FormRPartADto.java
```diff
@@ -33,78 +34,138 @@
...
+  @NotNull
   @Size(min = 1, max = 100)
   private String forename;
+
+  @NotNull
   @Size(min = 1, max = 100)
   private String surname;
+
+  @NotNull
   @Size(min = 1, max = 100)
```
john12321, line 65: FE has GMC maxLength = 20 ?

ReubenRobertsHEE: PartA - GMC number 100 max, PartB 20 max afaik

Judge40 (approved):
I didn't realise how messy the validation would be for those inter-dependant and mutually exclusive fields. Nice work.

One nit-pick would be some fields/method in the tests that could be made private, but I'm not going to block it for that.

PR #599, opened 2024-01-16
fix: preserve unsubmitted status in S3 when moving form

DorisWongNHS:
Set lifeCycleStatus according to S3 record when moving form as current "unsubmit" fnuction do not update lifeCycleStatus in DB


EdwardBarclay (approved):
LGTM

PR #604, opened 2024-01-24
fix: allow completion date to be null

ReubenRobertsHEE:
Also general fixes to better handle nulls in date annotations; prohibiting nulls is better handled with an explicit separate @NotNull annotation.

NO-TICKET

### DIFF
--- a/src/test/java/uk/nhs/hee/tis/trainee/forms/annotations/LegalAgeValidatorTest.java
+++ b/src/test/java/uk/nhs/hee/tis/trainee/forms/annotations/LegalAgeValidatorTest.java
@@ -1,0 +1,152 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright 2024 Crown Copyright (Health Education England)
+ *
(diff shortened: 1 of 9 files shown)


src/test/java/uk/nhs/hee/tis/trainee/forms/annotations/LegalAgeValidatorTest.java
```diff
@@ -0,0 +1,152 @@
...
+import static org.mockito.Mockito.when;
+
+import java.time.LocalDate;
+import javax.validation.ConstraintValidatorContext;
+import javax.validation.ConstraintValidatorContext.ConstraintViolationBuilder;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.mockito.Mock;
+
+class LegalAgeValidatorTest {
```
ReubenRobertsHEE, line 37: This is included because otherwise the null condition is never tested by the existing dto annotation validation tests, since it is masked by a @NotNull

john12321 (approved):

PR #606, opened 2024-01-30
feat: trim all form strings

ReubenRobertsHEE:
TIS21-5662

### DIFF
--- a/src/main/java/uk/nhs/hee/tis/trainee/forms/config/Config.java
+++ b/src/main/java/uk/nhs/hee/tis/trainee/forms/config/Config.java
@@ -25,6 +25,7 @@ import com.amazonaws.services.s3.AmazonS3;
 import com.amazonaws.services.s3.AmazonS3ClientBuilder;
 import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.module.SimpleModule;
 import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
 import java.text.SimpleDateFormat;
 import java.time.LocalDateTime;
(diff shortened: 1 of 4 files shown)


src/main/java/uk/nhs/hee/tis/trainee/forms/config/StringDeserializer.java
```diff
@@ -0,0 +1,14 @@
...
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import java.io.IOException;
+
+public class StringDeserializer extends JsonDeserializer<String> {
+
+  @Override
+  public String deserialize(JsonParser p, DeserializationContext ctxt) throws IOException {
+    return p.getValueAsString().trim();
```
Judge40, line 12: `String.strip()` covers some edge cases that we might want to cover off while we're at it.

Judge40 (approved):

src/main/java/uk/nhs/hee/tis/trainee/forms/config/Config.java
```diff
@@ -34,6 +35,8 @@
 @Configuration
 public class Config {
 
+  public static final String CUSTOM_STRING_DESERIALIZER_MODULE_NAME = "customStringDeserializer";
```
ReubenRobertsHEE, line 38: I agree giving the module a specific name seems unnecessary; the issue I had is that the test to see whether it's registered correctly wont work as it does for the JavaTimeModule. It gives 'SimpleModule-1' in getRegisteredModuleIds() if its not assigned a specific name - and 'SimpleModule-2' when creating a new one to compare it to, so the comparison fails 🤔

Judge40 (approved):

PR #639, opened 2024-06-04
fix: always get latest form contents

Judge40:
Changes made to unsubmitted forms are not stored back to S3 until re-submitted. When getting form contents the S3 contents are the default if it exists.
This leads to changes saved to the database being ignored when returning to edit the form further. Appearing, from a user point of view, that the updates were never saved.

Update the form services to check the modification time and always return the latest version of the form.

TIS21-6130

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -10,7 +10,7 @@ plugins {
 }
 
 group = "uk.nhs.hee.tis.trainee"
-version = "0.16.0"
+version = "0.16.1"
 
 configurations {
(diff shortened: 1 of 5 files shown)


Judge40:
I was originally checking for just `UNSUBMITTED` forms, grabbing the DB copy and overwriting the lifecycle state but went with this more generic approach based on the modification time.

EdwardBarclay (approved):

DorisWongNHS (approved):

PR #650, opened 2024-07-19
Add fields to forms to track linked programme membership and is arcp

EdwardBarclay:
TIS21-6221


john12321:
Fields look fine to me, Ed, but I will leave the review/approval to those BE Java folks ;)

PR #653, opened 2024-07-29
feat(formr): add linked PM and isARCP flag

Judge40:
Update Form R Part A and Part B to allow a programme membership ID to be stored against the form, this will show which PM the form is being submitted for. Also include a new `isArcp` flag so the user can specify whether their submission is towards ARCP or not.

TIS21-5611
TIS21-6280

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -10,7 +10,7 @@ plugins {
 }
 
 group = "uk.nhs.hee.tis.trainee"
-version = "0.17.0"
+version = "0.18.0"
 
 configurations {
   compileOnly {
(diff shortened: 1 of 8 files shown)


Judge40:
@john12321 this is relevant PR for the changes you're wanted, including the PM ID on the form list result.
The field name has changed from the original `linkedProgrammeUuid` to `programmeMembershipId`, so you'll need to make the equivalent frontend changes.

DorisWongNHS (approved):
LGTM

john12321:
> @john12321 this is relevant PR for the changes you're wanted, including the PM ID on the form list result.
> 
> The field name has changed from the original `linkedProgrammeUuid` to `programmeMembershipId`, so you'll need to make the equivalent frontend changes.

Cheers @Judge40. The FE commits are a real dogs dinner atm so that name change is the least of my worries!

PR #662, opened 2024-09-02
refactor: alter the expectedMetadata map for formr's to include isArc…

EdwardBarclay:
…p and programmeMembershipID

in order to make the isArcp flag and programmeMembershipId tied to each formr available on the FE of admins-ui these need to be within the custom-metadata. the Map.of was updated to a Map.ofEntries as there was a limit of 10 objects for Map.of which is increaded for Map.ofEntries and should have no affect on structure or utility

TICKET TIS21-6201

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -10,7 +10,7 @@ plugins {
 }
 
 group = "uk.nhs.hee.tis.trainee"
-version = "0.18.1"
+version = "0.18.3"
 
 configurations {
   compileOnly {
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -10,7 +10,7 @@ plugins {
 }
 
 group = "uk.nhs.hee.tis.trainee"
-version = "0.18.3"
+version = "0.19.0"
 
 configurations {
   compileOnly {
--- a/src/main/java/uk/nhs/hee/tis/trainee/forms/model/AbstractForm.java
+++ b/src/main/java/uk/nhs/hee/tis/trainee/forms/model/AbstractForm.java
@@ -37,9 +37,7 @@ public abstract class AbstractForm {
   @Indexed
   @Field(value = "traineeTisId")
   private String traineeTisId;
-  private Boolean isArcp;
 
-  private UUID programmeMembershipId;
   private LifecycleState lifecycleState;
   private LocalDateTime submissionDate;
   private LocalDateTime lastModifiedDate;
--- a/src/main/java/uk/nhs/hee/tis/trainee/forms/model/AbstractForm.java
+++ b/src/main/java/uk/nhs/hee/tis/trainee/forms/model/AbstractForm.java
@@ -37,7 +37,9 @@ public abstract class AbstractForm {
   @Indexed
   @Field(value = "traineeTisId")
   private String traineeTisId;
+  private Boolean isArcp;
 
+  private UUID programmeMembershipId;
   private LifecycleState lifecycleState;
   private LocalDateTime submissionDate;
   private LocalDateTime lastModifiedDate;
--- a/src/main/java/uk/nhs/hee/tis/trainee/forms/repository/AbstractCloudRepository.java
+++ b/src/main/java/uk/nhs/hee/tis/trainee/forms/repository/AbstractCloudRepository.java
@@ -20,6 +20,8 @@
 
 package uk.nhs.hee.tis.trainee.forms.repository;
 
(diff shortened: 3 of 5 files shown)


src/main/java/uk/nhs/hee/tis/trainee/forms/model/AbstractForm.java
```diff
@@ -37,7 +37,9 @@ public abstract class AbstractForm {
   @Indexed
   @Field(value = "traineeTisId")
   private String traineeTisId;
+  private Boolean isArcp;
 
+  private UUID programmeMembershipId;
```
Judge40, line 42: This class will be used for other forms, like LTFT, in the future which are not related to PMs or ARCPs.
You could create an intermediate `AbstractLinkedForm` between this abstract and and the two Form R concrete classes.

src/main/java/uk/nhs/hee/tis/trainee/forms/repository/AbstractCloudRepository.java
```diff
@@ -91,17 +93,19 @@ public T save(T form) {
...
-              form.getSubmissionDate().format(DateTimeFormatter.ISO_LOCAL_DATE_TIME),
-              "traineeid", form.getTraineeTisId(),
-              "deletetype", DeleteType.PARTIAL.name(),
-              "fixedfields", FIXED_FIELDS
+          .metadata(Map.ofEntries(
+              entry("id", form.getId().toString()),
+              entry("name", fileName),
+              entry("type", "json"),
+              entry("isarcp", form.getIsArcp().toString()),
+              entry("programmemembershipid", form.getProgrammeMembershipId().toString()),
```
Judge40, line 101: Why do we need this in the metadata?
The metadata should be generic for all forms, but these fields are only needed for Form Rs

Judge40 (changes requested):

src/main/java/uk/nhs/hee/tis/trainee/forms/repository/AbstractCloudRepository.java
EdwardBarclay: for it to be read in the formr view in admins, when speaking with Steve he said it would need to be in the custom metadata for the FE to be able to read it.

src/main/java/uk/nhs/hee/tis/trainee/forms/model/AbstractForm.java
Judge40: I think this comment is still valid, but given the different combinations of ARCP and which objects we might link forms to (I don't know whether we might need to link to placements in the future too) we probably need to keep these fields on the Form R classes instead of an abstract.

---

Using some interfaces might keep things more sensible, but it wouldn't change the data structure and the fields would still need to be declared on the FormRPartA/FormRPartB classes. So this could be something we do as part of the the LTFT work;

e.g.
```java
public interface ProgrammeMembershipLinked {
  public UUID getProgrammeMembershipId();
  public void setProgrammeMembershipId(UUID programmeMembershipId);
}

public interface ArcpLinked {
  public boolean getIsArcp();
  public void setIsArcp(boolean isArcp);
}

public class FormRPartA extends AbstractForm implements ProgrammeMembershipLinkedForm, ValidForArcpForm  {
  // Snip
  
  private UUID programmeMembershipId;
  private boolean isArcp;
   
  // Snip
}

public class FormRPartB extends AbstractForm implements ProgrammeMembershipLinkedForm, ValidForArcpForm  {
  // Snip
  
  private UUID programmeMembershipId;
  private boolean isArcp;
   
  // Snip
}

public class LessThanFullTime extends AbstractForm implements ProgrammeMembershipLinkedForm {
  // Snip
  
  private UUID programmeMembershipId;
   
  // Snip
}

```

Judge40 (changes requested):
I'm happy enough with the other changes but would prefer us to lift the `isArcp` and `programmeMembershipId` up to the PartA/PartB classes instead of the abstract.
I've given an example of what that might look like in the future, but you don't need to start creating interfaces right now unless you **really** want to 😅

build.gradle.kts
```diff
@@ -10,7 +10,7 @@ plugins {
 }
 
 group = "uk.nhs.hee.tis.trainee"
-version = "0.18.1"
```
Judge40, line 13: This is definitely more feature than bug

ReubenRobertsHEE (approved):

PR #664, opened 2024-10-08
feat: generate and upload PDF when COJ signed

Judge40:
When a COJ is signed and confirmation received back from TIS a PDF copy of it should be generated and uploaded to S3.
A notification should then be sent to SNS to inform other services that a PDF has been published for the COJ.

TIS21-6106
TIS21-6121


Judge40:
I have no intention to resolve the 7 code smells right now.

src/main/java/uk/nhs/hee/tis/trainee/forms/event/ConditionsOfJoiningSignedEvent.java
```diff
@@ -0,0 +1,45 @@
...
+ * @param traineeId             The trainee/person ID who signed the form.
+ * @param programmeMembershipId The programme membership that COJ relates to.
+ * @param programmeName         The name of the related programme.
+ * @param conditionsOfJoining   The Conditions of Joining detail.
+ */
+public record ConditionsOfJoiningSignedEvent(
+
+    @JsonAlias("personId")
+    String traineeId,
+    UUID programmeMembershipId,
```
ReubenRobertsHEE, line 40: Does this not arrive as "uuid" as per https://github.com/Health-Education-England/tis-trainee-sync/blob/958dd07fb543a2a9c304b4c93a039a399e582f4c/src/main/java/uk/nhs/hee/tis/trainee/sync/model/ProgrammeMembership.java#L40 ?

ReubenRobertsHEE: (Or are you planning to tweak trainee-sync to emit according to this structure?)

src/main/java/uk/nhs/hee/tis/trainee/forms/service/PdfPublisherService.java
```diff
@@ -0,0 +1,175 @@
...
+   * @param formType  The type of form being uploaded.
+   * @param filename  The filename without extension e.g. PM ID.
+   * @param pdf       The bytes of the PDF file.
+   * @return An object referencing the published PDF.
+   */
+  private PublishedPdf upload(String traineeId, String formType, String filename, byte[] pdf) {
+    log.info("Uploading generated {} for trainee '{}': {}.pdf", formType, filename, traineeId);
+    String key = String.format("%s/forms/%s/%s.pdf", traineeId, formType.toLowerCase(), filename);
+
+    s3Template.upload(publishBucket, key, new ByteArrayInputStream(pdf));
```
ReubenRobertsHEE, line 152: Are we ok doing this synchronously? That would match how we upload the form json, but these might be slightly larger files.

src/test/java/uk/nhs/hee/tis/trainee/forms/service/PdfPublisherServiceTest.java
```diff
@@ -0,0 +1,180 @@
...
+  private PdfPublisherService service;
+
+  private TemplateEngine templateEngine;
+  private S3Template s3Template;
+  private SnsTemplate snsTemplate;
+
+  @BeforeEach
+  void setUp() {
+    templateEngine = mock(TemplateEngine.class);
+    when(templateEngine.process(any(TemplateSpec.class), any())).thenReturn("<html></html>");
```
ReubenRobertsHEE, line 77: Aw, I was looking forward to seeing some beautiful end to end tests 😈

ReubenRobertsHEE:
All looks pretty solid to me, my only hesitation is the structure of the incoming COJ-signed events.

src/main/java/uk/nhs/hee/tis/trainee/forms/event/ConditionsOfJoiningSignedEvent.java
Judge40: You're right 😬 

I'll add an alias 🌮

src/main/java/uk/nhs/hee/tis/trainee/forms/service/PdfPublisherService.java
Judge40: A hadn't even considered that the S3Template wasn't using an async client under the hood 🤦🏻 
While I'm not sure there's a massive amount of benefit right now, we'll slowly be adding PDF uploads from Form Rs and LTFT so I'll just refactor it now.

ReubenRobertsHEE (approved):

src/test/java/uk/nhs/hee/tis/trainee/forms/service/PdfPublisherServiceTest.java
ReubenRobertsHEE: I'll expect it tomorrow then 👨‍🍼

PR #671, opened 2024-12-11
chore: local application property defaults

Judge40:
Create an `application-local.yml` file with sensible defaults for running this service locally.
The profile should be activated by setting
`SPRING_PROFILES_ACTIVE=local` in the env vars when running this service. No other env vars need to be set if everything is running directly on localhost, but `DB_HOST` and `LOCALSTACK_HOST` should still be set when running inside a container.
`SIGNATURE_SECRET_KEY` can still be used to ensure services all match, but a default value is provided.

There is still a pre-requisite that the queues and buckets listed in `application-local.yml` are created in the localstack instance.

NO-TICKET

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -10,7 +10,7 @@ plugins {
 }
 
 group = "uk.nhs.hee.tis.trainee"
-version = "0.21.1"
+version = "0.21.2"
 
 configurations {
   compileOnly {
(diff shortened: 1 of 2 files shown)


Judge40:
There is a small caveat, that this is intended to run with the latest version of localstack. It may not work as intended with the outdated version used by the handbook. Further updates will bring everything back in to line.

DorisWongNHS (approved):

PR #673, opened 2025-01-21
feat: get LTFT summaries api

ReubenRobertsHEE:
This includes a refactoring to replace the case-by-case checks on the authorization token with an interceptor.

Dependent on: API gateway (https://github.com/Health-Education-England/TIS-OPS/pull/912) 

TODO: refine LTFT form entity (and manually insert test data into staging collection?)

TIS21-6903


src/test/java/uk/nhs/hee/tis/trainee/forms/api/FormRPartAResourceTest.java
```diff
@@ -103,6 +107,7 @@ void setup() {
     FormRPartAResource formRPartAResource = new FormRPartAResource(service, validator);
     mockMvc = MockMvcBuilders.standaloneSetup(formRPartAResource)
         .setMessageConverters(jacksonMessageConverter)
+        .addInterceptors(new TraineeIdentityInterceptor(new TraineeIdentity()))
```
ReubenRobertsHEE, line 110: This feels quite hacky, but it was the only way I could get a the interceptor preHandle() to work.

src/test/java/uk/nhs/hee/tis/trainee/forms/api/FormRPartAResourceTest.java
```diff
@@ -130,7 +135,7 @@ void postShouldNotCreateFormWhenNoToken() throws Exception {
     mockMvc.perform(post("/api/formr-parta")
             .contentType(TestUtil.APPLICATION_JSON_UTF8)
             .content(TestUtil.convertObjectToJsonBytes(dto)))
-        .andExpect(status().isBadRequest());
+        .andExpect(status().isForbidden());
```
ReubenRobertsHEE, line 138: The change in response here and in other tests is due to the interceptor response. I _think_ this is not a (serious) breaking change.

src/main/java/uk/nhs/hee/tis/trainee/forms/api/ConditionsOfJoiningResource.java
```diff
@@ -49,9 +50,11 @@
 public class ConditionsOfJoiningResource {
 
   private final PdfService pdfService;
+  private final TraineeIdentity traineeIdentity;
 
-  public ConditionsOfJoiningResource(PdfService pdfService) {
+  public ConditionsOfJoiningResource(PdfService pdfService, TraineeIdentity traineeIdentity) {
```
ReubenRobertsHEE, line 55: The forms and LTFT resources have the trainee identity injected into the service, rather than into the resource. Since the CoJ doesn't really have a dedicated service, the trainee identity is injected directly into the resource, but this inconsistency might be worth ironing-out.

ReubenRobertsHEE:
Here's an example output from http://local.tis-selfservice.com/api/forms/ltft

```
[
{
"id": "679368af24cb6f4c285cd3ac",
"name": "Test form 1",
"programmeMembershipId": "c6cc64d6-11e8-45d8-91ae-b01bbf47cb06",
"status": "DRAFT",
"created": "2025-01-01T01:01:01.001Z",
"lastModified": "2025-01-02T02:02:02.002Z"
}
]
```

src/main/java/uk/nhs/hee/tis/trainee/forms/dto/LtftSummaryDto.java
```diff
@@ -0,0 +1,46 @@
...
+ * A summary of a LTFT application.
+ */
+@Builder
+public record LtftSummaryDto(
+    @JsonSerialize(using = ToStringSerializer.class)
+    ObjectId id,
+
+    String name,
+    UUID programmeMembershipId,
+    LifecycleState status,  //assumption: same as for FormR forms.
```
DorisWongNHS, line 42: Is it going to have the `APPROVED`, `WITHDRAWN` and `REJECTED` for LTFT status?

ReubenRobertsHEE: Yeah, that would make sense. I am not sure if we want to set up a new enum for LTFT states, or extend the existing one we use for FormRs.

src/main/java/uk/nhs/hee/tis/trainee/forms/model/LtftForm.java
```diff
@@ -0,0 +1,87 @@
...
+
+package uk.nhs.hee.tis.trainee.forms.model;
+
+import java.time.Instant;
+import java.time.LocalDate;
+import java.util.UUID;
+import lombok.Builder;
+import org.bson.types.ObjectId;
+import org.springframework.data.annotation.CreatedDate;
+import org.springframework.data.annotation.Id;
```
ReubenRobertsHEE, line 30: This is a bare-bones model. As only the summary is accessible (and there is no write functionality) it seemed ok to make this a placeholder until the full entity design is slightly more finalised.

src/main/java/uk/nhs/hee/tis/trainee/forms/model/LtftForm.java
```diff
@@ -0,0 +1,87 @@
...
+ * @param traineeId           The trainee ID.
+ * @param name                The form name.
+ * @param programmeMembership The form programme membership.
+ * @param status              The form status.
+ * @param created             When the form was created.
+ * @param lastModified        When the form was last updated.
+ */
+@Document("LtftForm")
+@Builder
+public record LtftForm(
```
ReubenRobertsHEE, line 50: This is a bare-bones model. As only the summary is accessible (and there is no write functionality) it seemed ok to make this a placeholder until the full entity design is slightly more finalised.

src/main/java/uk/nhs/hee/tis/trainee/forms/dto/LtftSummaryDto.java
Judge40: We probably want to extend the existing enum, but may need to add some validation for valid statuses per form type

src/test/java/uk/nhs/hee/tis/trainee/forms/api/FormRPartAResourceTest.java
Judge40: I've pushed an additional commit with some refactoring for this and the Part B equivalent.

src/test/java/uk/nhs/hee/tis/trainee/forms/api/FormRPartAResourceTest.java
Judge40: It's a more accurate response, so I say this is fine. Can't foresee it breaking anything, unless a third party is using our undocumented API 😉

src/main/java/uk/nhs/hee/tis/trainee/forms/api/ConditionsOfJoiningResource.java
Judge40: That's sensible enough to keep the scope of this PR from creeping too far. Do you want to put a TODO here for that refactoring so we can track it via Sonar?

src/main/java/uk/nhs/hee/tis/trainee/forms/interceptor/TraineeIdentityInterceptor.java
```diff
@@ -0,0 +1,87 @@
...
+   * @return True if it matches any API requiring a trainee ID, otherwise false.
+   */
+  private boolean apiRequiresTraineeId(String requestUri) {
+    for (String regexp : INTERCEPTOR_API_REGEXPS) {
+      if (requestUri.matches(regexp)) {
+        return true;
+      }
+    }
+    return false;
+  }
```
Judge40, line 86: This could be configured when registering the handler in `InterceptorConfiguration` using [addPathPatterns](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/servlet/config/annotation/InterceptorRegistration.html#addPathPatterns(java.lang.String...))

src/main/java/uk/nhs/hee/tis/trainee/forms/mapper/LtftMapper.java
```diff
@@ -0,0 +1,55 @@
...
+  @Mapping(target = "programmeMembershipId", source = "programmeMembership.id")
+  LtftSummaryDto toSummaryDto(LtftForm entity);
+
+  /**
+   * Convert a list of {@link LtftForm} to {@link LtftSummaryDto} DTOs.
+   *
+   * @param entities The entities to convert to DTOs.
+   * @return The equivalent summary DTOs.
+   */
+  @Mapping(target = "programmeMembershipId", source = "programmeMembership.id")
```
Judge40, line 53: This mapping is redundant since it will call `toSummaryDto` multiple times anyway.

src/main/java/uk/nhs/hee/tis/trainee/forms/dto/LtftSummaryDto.java
ReubenRobertsHEE: Happy to build that in now, unless we'd prefer to keep this PR from getting any bigger?

src/main/java/uk/nhs/hee/tis/trainee/forms/api/LtftResource.java
```diff
@@ -0,0 +1,65 @@
...
+  public LtftResource(LtftService service) {
+    this.service = service;
+  }
+
+  /**
+   * Retrieve a list of LTFT summaries for the logged-in user.
+   *
+   * @return The list of LTFT summaries, or an empty list if none found.
+   */
+  @GetMapping("/ltft")
```
Judge40, line 59: Move the `/ltft` path part to the class level

src/main/java/uk/nhs/hee/tis/trainee/forms/config/InterceptorConfiguration.java
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Configuration for interceptors.
+ */
+@Configuration
+public class InterceptorConfiguration implements WebMvcConfigurer {
+
+  @Override
+  public void addInterceptors(InterceptorRegistry registry) {
+    registry.addInterceptor(traineeIdentityInterceptor());
```
Judge40, line 40: As mentioned elsewhere, we should be able to add the path filtering here.

src/main/java/uk/nhs/hee/tis/trainee/forms/service/FormRPartAService.java
```diff
@@ -73,15 +76,27 @@ public class FormRPartAService {
...
+    this.traineeIdentity = traineeIdentity;
+  }
+
+  /**
+   * Get the logged-in trainee's ID.
+   *
+   * @return The trainee TIS ID.
+   */
+  public String getLoggedInTraineeId() {
+    return traineeIdentity.getTraineeId();
```
Judge40, line 99: There are a couple of better options for this
1. Inject the TraineeIdentity in to the resource as well the as the service, so we don't need to retrieve it separately.
2. Mimic trainee detail's [NotRecordOwnerException ](https://github.com/Health-Education-England/tis-trainee-details/blob/c139b01f5f3aa653df7d59ab5cec325cb1c6993f/src/main/java/uk/nhs/hee/trainee/details/exception/NotRecordOwnerException.java)and move the ID verification to the service class.

src/main/java/uk/nhs/hee/tis/trainee/forms/service/FormRPartBService.java
```diff
@@ -75,11 +78,21 @@ public class FormRPartBService {
...
+    this.traineeIdentity = traineeIdentity;
+  }
+
+  /**
+   * Get the logged-in trainee's ID.
+   *
+   * @return The trainee TIS ID.
+   */
+  public String getLoggedInTraineeId() {
+    return traineeIdentity.getTraineeId();
```
Judge40, line 95: As above.

Judge40 (changes requested):

src/main/java/uk/nhs/hee/tis/trainee/forms/dto/LtftSummaryDto.java
Judge40: I don't think we need to do that just yet, we can add those statuses as part of the appropriate tickets that will enable that functionality.

ReubenRobertsHEE:
@Judge40 I didn't notice your push https://github.com/Health-Education-England/tis-trainee-forms/pull/673/commits/7bfc5a8f5b8d58bac43a1c4818e6139e2b820b21 before I made the various PR-comment changes, hence the cherry-pick. It looked like a great simplification of the tests, but my changes broke it. The mockMvc seems to have reverted to not having the interceptor properly included, because even passing tests are not necessarily valid (e.g. removing the authorization header, postShouldCreateFormWhenDtoValidationPasses() still 'passes').

I'll keep looking for a non-hand-cranked fix, but if there's anything obvious that jumps out at you, could save some time....

src/main/java/uk/nhs/hee/tis/trainee/forms/config/InterceptorConfiguration.java
```diff
@@ -35,9 +36,19 @@
 @Configuration
 public class InterceptorConfiguration implements WebMvcConfigurer {
 
+  // Endpoints are a mix of authenticated (public) and unauthenticated (internal), limit
+  // trainee ID verification to LTFT, COJ and FormR endpoints for now.
+  protected static final String[] TRAINEE_ID_APIS
+      = {"/api/coj",
+        "/api/formr-parta", "/api/formr-partas", "/api/formr-parta/**",
+        "/api/formr-partb", "/api/formr-partbs", "/api/formr-partb/**",
+        "/api/ltft", "/api/ltft/**"};
```
Judge40, line 45: These are a bit uglier than I was hoping for, but I guess we don't have the same flexibility.

Judge40 (approved):

PR #676, opened 2025-02-03
fix: index forms collections

ReubenRobertsHEE:
Responses to the get formr-part[a|b]s has strted getting a bit sluggish.

NO-TICKET (TIS21-6954)

### DIFF
--- a/src/main/java/uk/nhs/hee/tis/trainee/forms/api/FormRPartAResource.java
+++ b/src/main/java/uk/nhs/hee/tis/trainee/forms/api/FormRPartAResource.java
@@ -82,7 +82,7 @@ public class FormRPartAResource {
   @PostMapping("/formr-parta")
   public ResponseEntity<FormRPartADto> createFormRPartA(@RequestBody FormRPartADto dto)
       throws URISyntaxException, MethodArgumentNotValidException {
-    log.debug("REST request to save FormRPartA : {}", dto);
+    log.info("REST request to save FormRPartA : {}", dto);
     if (dto.getId() != null) {
       return ResponseEntity.badRequest().headers(HeaderUtil
           .createFailureAlert(ENTITY_NAME, "idexists",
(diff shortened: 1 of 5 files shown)


src/main/java/uk/nhs/hee/tis/trainee/forms/api/FormRPartAResource.java
```diff
@@ -82,7 +82,7 @@ public FormRPartAResource(FormRPartAService service, FormRPartAValidator validat
   @PostMapping("/formr-parta")
   public ResponseEntity<FormRPartADto> createFormRPartA(@RequestBody FormRPartADto dto)
       throws URISyntaxException, MethodArgumentNotValidException {
-    log.debug("REST request to save FormRPartA : {}", dto);
+    log.info("REST request to save FormRPartA : {}", dto);
```
ReubenRobertsHEE, line 85: I've made these all 'info' level logging because that's what we generally use, and it makes digging through the historic logs a bit easier to find what happened to particular user's forms. 'Debug' would only help with debugging a particular issue if we deployed specifically with that logging level, i.e. it wouldn't help diagnostics for past events.

But I could be wrong 😁

src/main/java/uk/nhs/hee/tis/trainee/forms/config/MongoConfiguration.java
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  /**
+   * Add custom indexes to the Mongo collections.
+   */
+  @PostConstruct
+  public void initIndexes() {
+    IndexOperations formRpartAindexOps = template.indexOps(FormRPartA.class);
+    formRpartAindexOps.ensureIndex(new Index().on("traineeTisId", Sort.Direction.ASC));
+    formRpartAindexOps.ensureIndex(new Index().on("lifecycleState", Sort.Direction.ASC));
```
ReubenRobertsHEE, line 52: Two distinct indexes instead of one compound one might give us a bit more flexibility, but it was a toss-up.

Judge40 (approved):
LGTM

PR #678, opened 2025-02-04
feat: add LTFT count endpoint for admins

Judge40:
Add an endpoint `/api/admin/ltft/count` which will return a count of LTFT applications that are associated with an admin's local office. The result should be filterable on status, but passing a URL parameter such `status=UNSUBMITTED,SUBMITTED`.

Local office filtering will follow, only status filtering will be supported at this point.

Refactor the Gradle build script to create an integration test suite separate from the existing test suite. This integration suite will make full use of test containers to allow testing of several layers, from API, through service, to repository.
Existing integration tests have not been moved yet.

TIS21-6599
TIS21-6938

### DIFF
--- a/src/test/java/uk/nhs/hee/tis/trainee/forms/api/AdminLtftResourceTest.java
+++ b/src/test/java/uk/nhs/hee/tis/trainee/forms/api/AdminLtftResourceTest.java
@@ -1,0 +1,79 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright 2025 Crown Copyright (Health Education England)
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of this software and
+ * associated documentation files (the "Software"), to deal in the Software without restriction,
+ * including without limitation the rights to use, copy, modify, merge, publish, distribute,
+ * sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all copies or
+ * substantial portions of the Software.
+ *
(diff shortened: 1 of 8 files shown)


Judge40:
This doesn't solve the whole problem, but it's already chunky due to the build script changes. Since the remaining LO filtering will take quite a bit of work I'd like to get this initial piece merged in so it can be used for local FE development.

src/test/java/uk/nhs/hee/tis/trainee/forms/api/AdminLtftResourceTest.java
```diff
@@ -0,0 +1,79 @@
...
+    assertThat("Unexpected filter states.", capturedStates, hasItems(UNSUBMITTED, SUBMITTED));
+  }
+
+  @Test
+  void shouldGetCountResponse() {
+    when(service.getAdminLtftCount(any())).thenReturn(40L);
+
+    ResponseEntity<String> response = controller.getAdminLtftCount(Set.of());
+
+    assertThat("Unexpected response code.", response.getStatusCode().value(), is(200));
```
ReubenRobertsHEE, line 75: Not a fan of `getStatusCode(), is(OK)` ? 😁

ReubenRobertsHEE (dismissed):
LGTM

src/test/java/uk/nhs/hee/tis/trainee/forms/api/AdminLtftResourceTest.java
Judge40: Cleaned that last bit up if you want to re-review @ReubenRobertsHEE

ReubenRobertsHEE (approved):

PR #680, opened 2025-02-06
refactor: form inheritance

Judge40:
Refactor the AbstractForm, Form Rs and LtftForm in to a shared, more consistent hierarchy.
Since LTFT will have a state audit history included the structure for lifecycle state and submission date will change, so those should be moved to an abstract class shared by just the existing forms (FormR Part A and B). Additionally the lastModified field will be an Instant instead of LocalDateTime, so that field should also be moved so it is only used by FormRs in its current form.

The LTFT form has been refactored to follow the conventions set by FormR, so the builder has been removed until the forms can be refactored again in sync.

NO-TICKET

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -10,7 +10,7 @@ plugins {
 }
 
 group = "uk.nhs.hee.tis.trainee"
-version = "0.23.0"
+version = "0.23.1"
 
 configurations {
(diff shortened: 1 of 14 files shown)


ReubenRobertsHEE (approved):
Looks reasonable!

PR #685, opened 2025-02-14
fix: set created date when updating form

ReubenRobertsHEE:
HACK

Without success I've tried the suggested workarounds on https://github.com/spring-projects/spring-data-rest/issues/1565  (`@JsonIgnore` or `@JsonProperty(access = JsonProperty.Access.READ_ONLY)` on the created field. Making the field final introduced other issues with inheritance, so I didn't pursue it. Using `@Version` on a new field and changing the `isNew()` code to refer to this made no difference either. 

(We don't actually use spring-data-rest specifically, but it seemed the clearest explanation of the issue from much googling around).

Heroic efforts by the author here: https://stackoverflow.com/questions/68441685/spring-data-mongodb-how-to-avoid-createdby-and-createddate-fields-update - but I'm not sure how much we want to pervert the MongoTemplate to our own base desires.

Likewise, using spring-data-jdbc might offer some slightly lower-level interventions (e.g. `@InsertOnlyProperty`) but I don't know if we want to go there. It seems we might still have to better control whether mongo uses full updates / partial updates (if I'm understanding the comments on https://stackoverflow.com/questions/73051733/createddate-set-to-null-on-update-spring-data-jdbc correctly).

TIS21-6792

### DIFF
--- a/src/main/java/uk/nhs/hee/tis/trainee/forms/service/LtftService.java
+++ b/src/main/java/uk/nhs/hee/tis/trainee/forms/service/LtftService.java
@@ -156,6 +156,7 @@ public class LtftService {
           formId, traineeId);
       return Optional.empty();
     }
+    form.setCreated(existingForm.get().getCreated()); //explicitly set otherwise form saved as 'new'
     LtftForm savedForm = ltftFormRepository.save(form);
     return Optional.of(mapper.toDto(savedForm));
   }
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -10,7 +10,7 @@ plugins {
 }
 
 group = "uk.nhs.hee.tis.trainee"
-version = "0.24.1"
+version = "0.24.2"
 
 configurations {
   compileOnly {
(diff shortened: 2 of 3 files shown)


src/main/java/uk/nhs/hee/tis/trainee/forms/service/LtftService.java
```diff
@@ -156,6 +156,7 @@ public Optional<LtftFormDto> updateLtftForm(UUID formId, LtftFormDto dto) {
           formId, traineeId);
       return Optional.empty();
     }
+    form.setCreated(existingForm.get().getCreated()); //explicitly set otherwise form saved as 'new'
```
ReubenRobertsHEE, line 159: hackitty hackitty hack!

If the 'isNew()' logic uses a version field, then you don't need to set the created date, but (a) you do need to set the version from the saved form and (b) created is set to null.

Judge40: For want of a better approach, this looks okay. Googling around seems like there's no simple/OOTB solution for it, besides partial updates (which may be easier to implement with the new self-contained `content`, but something to pursue later).

Judge40 (approved):

PR #686, opened 2025-02-14
refactir: identity interceptor

Judge40:
Currently only a TraineeIdentity and associated interceptor exists, however an Admin equivalent is now needed for TIS Admins.

Refactor the existing identity functionality to be more generic. Create an AdminIdentity and register a compatible interceptor, the identity should have `email`, `name` and `group` attributes.

TIS21-6599
TIS21-6938


ReubenRobertsHEE:
I'm still working through the PR, but I was thinking about the size of the JWT token and the groups list... and then realising I'm not 100% sure what groups these refer to. Are they cognito groups on the TIS admin side that encompass the roles a TIS user has assigned? They can't be TSS cognito groups, if they are controlling what a TIS user sees, right? Friday PM confusion 😩

Judge40:
> I'm still working through the PR, but I was thinking about the size of the JWT token and the groups list... and then realising I'm not 100% sure what groups these refer to. Are they cognito groups on the TIS admin side that encompass the roles a TIS user has assigned? They can't be TSS cognito groups, if they are controlling what a TIS user sees, right? Friday PM confusion 😩
> 
> Edit: ok, I see it on the other ticket: "The DBCs are set as a Cognito group and added to the auth token as cognito:groups."

Yeah, that it.
They aren't actually real tangible groups in Cognito, but when the token is generated the DBCs are injected in to the `cognito:groups` array.

ReubenRobertsHEE:
> > I'm still working through the PR, but I was thinking about the size of the JWT token and the groups list... and then realising I'm not 100% sure what groups these refer to. Are they cognito groups on the TIS admin side that encompass the roles a TIS user has assigned? They can't be TSS cognito groups, if they are controlling what a TIS user sees, right? Friday PM confusion 😩
> > Edit: ok, I see it on the other ticket: "The DBCs are set as a Cognito group and added to the auth token as cognito:groups."
> 
> Yeah, that's it. They aren't actually real tangible groups in Cognito, but when the TIS/Admins UI token is generated the DBCs are injected in to the `cognito:groups` array.

Cool cool. There are some users with ~100 DBCs. I'm not familiar with JWT compression, but I presume that is not (yet) _too_ big to be adding to every request!

Judge40:
> Cool cool. There are a few users with ~100 DBCs - but very few of them. Most have ~10 or less.
> 
> (Edit: a token for a user like that looks to be ~1.7kB which is maybe too big, but it seems safe to ignore the issue at this point)

I'd question the validity of those as genuine users, rather than miscofiguration or test/machine accounts.
Either way, I think we have 8kb to play with so we should be okay.

ReubenRobertsHEE (approved):

ReubenRobertsHEE:
I got a bit confused about the googled-guidance on max size I was getting, whether it was 8 kilobytes or 8 kilobits 😂

PR #687, opened 2025-02-14
feat: filter admin LTFT count by DBCs

Judge40:
Filter the result of the admin LTFT count endpoint based on the DBCs that the admin user is assigned to.
The DBCs are set as a Cognito group and added to the auth token as `cognito:groups`.
This may include non-DBC groups at some point, but those should never match a PM's DBC anyway.

TIS21-6599
TIS21-6938

### DIFF
--- a/src/main/java/uk/nhs/hee/tis/trainee/forms/service/LtftService.java
+++ b/src/main/java/uk/nhs/hee/tis/trainee/forms/service/LtftService.java
@@ -31,6 +31,7 @@ import org.springframework.stereotype.Service;
 import uk.nhs.hee.tis.trainee.forms.dto.LtftFormDto;
 import uk.nhs.hee.tis.trainee.forms.dto.LtftSummaryDto;
 import uk.nhs.hee.tis.trainee.forms.dto.enumeration.LifecycleState;
+import uk.nhs.hee.tis.trainee.forms.dto.identity.AdminIdentity;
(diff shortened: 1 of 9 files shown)


Judge40:
Raised a draft as there is a dependency on #686 and I don't want to add any more to that PR as it's already big. Besides any changes to #686 this PR **can** be reviewed now ready to merge once the dependency has merged.

src/main/java/uk/nhs/hee/tis/trainee/forms/service/LtftService.java
```diff
@@ -44,19 +45,22 @@
...
    *
-   * @param traineeIdentity    The logged-in trainee.
+   * @param adminIdentity      The logged-in admin, for admin features.
+   * @param traineeIdentity    The logged-in trainee, for trainee features.
    * @param ltftFormRepository The LTFT repository.
    * @param mapper             The LTFT mapper.
    */
-  public LtftService(TraineeIdentity traineeIdentity, LtftFormRepository ltftFormRepository,
-      LtftMapper mapper) {
+  public LtftService(AdminIdentity adminIdentity, TraineeIdentity traineeIdentity,
```
Judge40, line 61: I'm not overly keen on having two different identity fields in the same service, it feels a bit like setting a trap for someone to fall in to later.

Once we add status change endpoints it will make more sense to split this service and create an abstract for the shared functionality like status updates.

EdwardBarclay (approved):
LGTM

PR #689, opened 2025-02-20
feat(ltft): populate missing admin summary fields

Judge40:
Several fields in the LTFT Admin summary use dummy/default values, update the mappers to populate the fields that are currently possible.

TIS21-6600
TIS21-6940


src/main/java/uk/nhs/hee/tis/trainee/forms/api/AdminLtftResource.java
```diff
@@ -77,7 +77,8 @@ public ResponseEntity<String> getAdminLtftCount(
   @GetMapping
   ResponseEntity<PagedModel<LtftAdminSummaryDto>> getLtftAdminSummaries(
       @RequestParam(value = "status", required = false) Set<LifecycleState> states,
-      @PageableDefault(size = Integer.MAX_VALUE, sort = "created", direction = Direction.ASC)
+      @PageableDefault(size = Integer.MAX_VALUE, sort = "content.change.startDate",
+          direction = Direction.ASC)
```
ReubenRobertsHEE, line 81: I guess it would be good to confirm with the users, but I would expect a descending order sort to possibly be more useful here.

src/main/java/uk/nhs/hee/tis/trainee/forms/mapper/LtftMapper.java
```diff
@@ -21,23 +21,42 @@
...
  */
-@Mapper(componentModel = SPRING, uses = TemporalMapper.class)
-public interface LtftMapper {
+@Mapper(componentModel = SPRING, uses = TemporalMapper.class, injectionStrategy = CONSTRUCTOR)
+public abstract class LtftMapper {
+
+  private static final int NOTICE_PERIOD_DAYS = 112; // 16 weeks.
+
+  @Getter
+  @Setter(onMethod_ = @Autowired)
```
ReubenRobertsHEE, line 58: Google incoming... 😂

ReubenRobertsHEE (approved):

ReubenRobertsHEE:
(You'll be happy to know your job looks safe, for now... 😈 . Cursor confidently rewrote that mapper and associated tests, and broke it quite nicely.)

src/main/java/uk/nhs/hee/tis/trainee/forms/api/AdminLtftResource.java
Judge40: Unless I've got my thinking backwards, this is closest start date first - which feels like what we'd want?

src/main/java/uk/nhs/hee/tis/trainee/forms/mapper/LtftMapper.java
Judge40: I was hoping to be able to have an abstract `getTemporalMapper()` which would return the `TemporalMapper` from the generated `LtftMapperImpl`, but unfortunately mapstruct doesn't support that. So this is a bit of lombok magic to inject it here as well as in to the constructor on the implementation.

It does lead to some awkward initialisation in tests, but it is made a little easier with the `CONSTRUCTOR` injection strategy avoiding our previous reflection-based approach to initialising nested mappers.

src/main/java/uk/nhs/hee/tis/trainee/forms/api/AdminLtftResource.java
ReubenRobertsHEE: Ah right you are. I was thinking of all the past ones appearing at the top, but I presume those would be filtered out.

PR #692, opened 2025-02-25
feat(ltft): add admin LTFT status update endpoints

Judge40:
Add admin-facing endpoints for approval, unsubmit and withdraw state transitions.
The transition path will be validation and an error return if the transition is not allowed (e.g. DRAFT -> APPROVED).

TIS21-6540


Judge40:
Merged in to other branches, so no longer needed.

PR #694, opened 2025-02-26
feat(ltft): add missing fields to LTFT detail DTO

Judge40:
Refactor `LtftFormDto` and add the missing fields. The detail DTO has been based on the database model, with some additional fields taken from the API docs.

The current docs and the DTO are not an exact match, the DTO may need further tweaks in some places but the API docs are also known to be out of date.
Health-Education-England/tis-trainee-api#28 exists with some of the changes, already. Other changes like the `formRef` field still need to be added once PR-28 has been approved/merged.

The CCT date is calculated dynamically when converting from the database entity, the value will always be null until the calculation is implemented.

TIS21-6540

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -10,7 +10,7 @@ plugins {
 }
 
 group = "uk.nhs.hee.tis.trainee"
-version = "0.29.1"
+version = "0.29.2"
 
 configurations {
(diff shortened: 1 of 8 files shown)


ReubenRobertsHEE (approved):
LGTM.

PR #696, opened 2025-02-28
fix(ltft): allow nullable fields and don't null CCT Date

Judge40:
# fix(ltft): don't set null fields to default values

Several LTFT fields can have valid `null` values when working with a DRAFT form, if that section has not been filled in. Currently several fields are using primitive boolean and numerical types, which are being defaulted to `false`/`0`.

Update the LTFT entity and model to use non-primative types which can be defaulted to `null`.
Several of these fields should never be `null`, such as PM WTE which is always expected to be populated. However, defaulting to erroneous values causes more issues than missing data, as the source of the value becomes unknown/lost.

---

# fix(ltft): save given CCT Date
The current DTO includes CCT Date in the programme membership, but the
database model does not. The result is that the CCT Date is dropped and
cannot be retrieved again.

Add `cctDate` to the `CctChange` model and configure the `LtftMapper` to
copy the value across.

---

TIS21-6540

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -10,7 +10,7 @@ plugins {
 }
 
 group = "uk.nhs.hee.tis.trainee"
-version = "0.31.0"
+version = "0.31.1"
 
 configurations {
   compileOnly {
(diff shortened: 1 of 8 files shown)


ReubenRobertsHEE (approved):
Probably never seen so many null assertions in one place. Hope never to again 😁

Judge40:
> Probably never seen so many null assertions in one place. Hope never to again 😁

It feels a bit silly to do it in unit tests for the service, but it works for now.
They can be refactored to an integration test that actually verify the JSON response, which will be a lot more intuitive and cleaner.

PR #700, opened 2025-03-12
feat(ltft): sorting and filtering admin endpoints

Judge40:
The sorting and filtering of admin endpoints is currently limited as the internal field names are required.
As the internal field names are not transparent to the client, it should be possible to sort and filter based on the DTO field names.

Update LtftService to translate the supported field names for both sorting and filtering. The user filters should be added on top of the filters for DBC and non-DRAFT LTFTs, but not replace them.

To enable sorting on `submissionDate`, which is a calculated field, it should be made in to an actual field on the database model. The `submitted` timestamp will be updated each time the LTFT is submitted. Refactor AbstractAuditedForm to add the new field and remove the existing `getSubmitted()` method.

Update `LtftAdminSummaryDto` to include the missing `formRef` field.

Update `LtftContent` to add indexes for the filterable fields. `assignedAdmin.name` and `assignedAdmin.email` will not yet have indexes due to the shared `Person` type. Indexes will need to be created programmatically using `indexOps`.

TIS21-6934
TIS21-7098
TIS21-7099
TIS21-6943

### DIFF
--- a/src/main/java/uk/nhs/hee/tis/trainee/forms/model/AbstractAuditedForm.java
+++ b/src/main/java/uk/nhs/hee/tis/trainee/forms/model/AbstractAuditedForm.java
@@ -21,9 +21,9 @@
 
 package uk.nhs.hee.tis.trainee.forms.model;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import java.time.Instant;
 import java.util.ArrayList;
-import java.util.Comparator;
 import java.util.List;
 import java.util.UUID;
 import lombok.Builder;
@@ -69,6 +69,7 @@ public abstract class AbstractAuditedForm<T extends FormContent> extends Abstrac
    * @return The lifecycle state, or null if no lifecycle states exist.
    */
   @Override
+  @JsonIgnore
   public LifecycleState getLifecycleState() {
     if (status == null || status.current == null) {
       return null;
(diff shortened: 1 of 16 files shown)


src/main/java/uk/nhs/hee/tis/trainee/forms/model/AbstractAuditedForm.java
```diff
@@ -107,47 +108,41 @@ public void setLifecycleState(LifecycleState lifecycleState, Status.StatusDetail
         .revision(revision)
         .build();
 
+    Instant submitted;
+
+    if (lifecycleState == LifecycleState.SUBMITTED) {
+      submitted = statusInfo.timestamp;
+    } else {
+      submitted = status != null ? status.submitted : null;
```
ReubenRobertsHEE, line 116: So we preserve a previous `submitted` timestamp value if a form is subsequently unsubmitted/rejected, and only update `submitted` when the form is resubmitted?

Judge40: Yes, it's meant mostly as an indication when a form became "ready" to reivew/approve I guess. The original submission timestamp isn't as useful as knowing when it was submitted "correctly"

ReubenRobertsHEE (approved):

PR #701, opened 2025-03-18
feat(ltft): admin details PDF

Judge40:
Add an endpoint to allow TIS admins to download LTFT details as a PDF. The API path is the same as getting the JSON detail, but the `Accept` header should be set to `application/pdf`.

The PDF template is rough but data-wise complete. Further restructuring of the HTML and CSS styling will be needed.

TIS21-6635
TIS21-7177

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -10,7 +10,7 @@ plugins {
 }
 
 group = "uk.nhs.hee.tis.trainee"
-version = "0.35.0"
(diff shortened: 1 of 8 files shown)


Judge40:
Sample PDF: [admin-ltft-detail.pdf](https://github.com/user-attachments/files/19324609/admin-ltft-detail.pdf)

PR #702, opened 2025-03-20
Feat/post submit ltft

DorisWongNHS:
- New POST endpoint `/api/ltft/submit` to submit new LTFT form that have not been saved before
- Update PUT endpoint `/api/ltft/{formId}/submit` to save the DTO before changing into SUBMITTED status. To fix the issue when part of the form was not saved, the changes would loss if the PUT submit only updated the status according to the formId.

TIS21-7057
TIS21-7119

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -10,7 +10,7 @@ plugins {
 }
 
 group = "uk.nhs.hee.tis.trainee"
-version = "0.35.0"
+version = "0.36.0"
 
 configurations {
   compileOnly {
--- a/src/integrationTest/java/uk/nhs/hee/tis/trainee/forms/api/LtftResourceIntegrationTest.java
+++ b/src/integrationTest/java/uk/nhs/hee/tis/trainee/forms/api/LtftResourceIntegrationTest.java
@@ -65,6 +65,10 @@ import org.testcontainers.junit.jupiter.Testcontainers;
 import uk.nhs.hee.tis.trainee.forms.DockerImageNames;
 import uk.nhs.hee.tis.trainee.forms.TestJwtUtil;
 import uk.nhs.hee.tis.trainee.forms.dto.LtftFormDto;
+import uk.nhs.hee.tis.trainee.forms.dto.LtftFormDto.DeclarationsDto;
+import uk.nhs.hee.tis.trainee.forms.dto.LtftFormDto.StatusDto;
+import uk.nhs.hee.tis.trainee.forms.dto.LtftFormDto.StatusDto.LftfStatusInfoDetailDto;
+import uk.nhs.hee.tis.trainee.forms.dto.LtftFormDto.StatusDto.StatusInfoDto;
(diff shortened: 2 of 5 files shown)


ReubenRobertsHEE:
I wonder if instead of changing the submit PUT endpoint and making it different to the other state-change endpoints, we should be handling this in the FE to manage the process of creating -> saving -> submitting by e.g. disabling the 'Submit' button until the form has been saved? @john12321 what do you think?

ReubenRobertsHEE:
The other reason I'm not super keen on the direct create -> submit POST endpoint, is that it is possible that a form can be successfully saved, but not submitted (e.g. if it's missing some details or otherwise is invalid). It feels like communicating the range of possible states back to the FE might become a bit fiddly for the FE to interpret:
1. Saved ok, submitted ok, here's your submitted form DTO with brand-new formId
2. Saved ok, couldn't be submitted, at the moment we just return a 'bad request' which could have the reasons the form is not submittable in the body error object but not the new formId afaik, so how would the FE know that it now needs to use the PUT endpoint next time? If it doesn't, then the trainee ends up with multiple copies of the same form because the next time it is POST'd it creates a new form in the back-end.
3. Didn't save ok (wrong trainee, invalid DTO, server error, etc.) again 'bad request' with reasons in the body error object

john12321:
> The other reason I'm not super keen on the direct create -> submit POST endpoint, is that it is possible that a form can be successfully saved, but not submitted (e.g. if it's missing some details or otherwise is invalid). It feels like communicating the range of possible states back to the FE might become a bit fiddly for the FE to interpret:
> 
> 1. Saved ok, submitted ok, here's your submitted form DTO with brand-new formId
> 2. Saved ok, couldn't be submitted, at the moment we just return a 'bad request' which could have the reasons the form is not submittable in the body error object but not the new formId afaik, so how would the FE know that it now needs to use the PUT endpoint next time? If it doesn't, then the trainee ends up with multiple copies of the same form because the next time it is POST'd it creates a new form in the back-end.
> 3. Didn't save ok (wrong trainee, invalid DTO, server error, etc.) again 'bad request' with reasons in the body error object

I think we need a chat! I might be missing something obvious here.
e.g. point 2. Saved OK (via autosave) will return the id which is stored and used in the PUT submit payload.

ReubenRobertsHEE:
Suggested a huddle. The context I was assuming is that the form has never been saved before, so the new POST submit endpoint was going to need to do that first save and submit all in one step, but if the submit part of that failed you'd just get a 'bad request' response to the FE.

DorisWongNHS:
As we suggested in the huddle, we will refactor the frontend to:
1. put the naming step in the declaration section
2. save when "submit" button is clicked on form preview page
3. submit when "confirm" button is clicked in the modal
So it will separate the request with each user button click, and make sure the specific results are returned to frontend for each actions.
Closing this PR as no backend change is needed.

PR #704, opened 2025-03-26
style: improve the LTFT PDF styling

Judge40:
Iterate on the LTFT PDF styling to produce a more sensible PDF, using the latest versions of `nhsuk` frontend CSS, `JSoup` and `OpenHtmlToPdf`.

Due to apparent limitations with the `min-width` media query i.e. `@media (min-width:123em)` used in the mobile-first `nhsuk` CSS library, manual overrides should be used in `nhsuk-9.3.0.override.css`. When evaluating the CSS the viewport size is used to match against `min-width`, however when generating the PDF the minimum width is never met. Several attempts were made to set content and page size before giving up and manually overriding with the expected styles.

Several fields on the PDF have been rearranged or combined to give a more sensible output.

TIS21-6635
TIS21-7117


Judge40:
Updated example: [admin-ltft-detail.pdf](https://github.com/user-attachments/files/19458630/admin-ltft-detail.pdf)  
There is a separate subtask to outline how the PDF should look, this current iteration meets the output of that sub-task as it stands.

There are some more minor things that could be considered in future iterations once more direction is provided
 - Nicer display of boolean fields (fontawesome, or just better wording than `true` `false`
 - Review of the included fields and strip back to what is important
 - NHS branding

src/main/resources/templates/ltft/admin.html
```diff
@@ -3,151 +3,180 @@
...
+            <div class="nhsuk-summary-list__row">
+              <dt class="nhsuk-summary-list__key">Programme end date</dt>
+              <dd class="nhsuk-summary-list__value" th:text="${#temporals.format(change?.cctDate, 'dd MMMM yyyy')} ?: _">Not Given</dd>
+            </div>
+          </dl>
+        </div>
+      </div>
+      <div class="nhsuk-card">
+        <div class="nhsuk-card__content">
+          <h2 class="nhsuk-card__heading">Reasons</h2>
```
ReubenRobertsHEE, line 122: Is it the th:block stuff which is throwing the column widths out? Its not a big deal at this point, but it looks odd that the list key entries are so much wider than they need to be in the Discussions and Reasons sections

Judge40: Possibly 🤷‍♂️
`th:block` elements disappear, but i think there is some weirdness with the whitespace left over from formatting them.

I'll check what happens if I one-liner it all 😅

src/integrationTest/java/uk/nhs/hee/tis/trainee/forms/service/PdfServiceIntegrationTest.java
```diff
@@ -0,0 +1,209 @@
...
+
+  /**
+   * Compare the bytes of a generated PDF against an existing example.
+   *
+   * @param resource          The filename of the existing PDF to check against, without extension.
+   * @param generatedPdfBytes The generated bytes.
+   * @return The number of discrepancies between the generation and expected results.
+   * @throws IOException If output files could not be created, or reading the bytes fails.
+   */
+  private int compareGeneratedPdf(String resource, byte[] generatedPdfBytes) throws IOException {
```
ReubenRobertsHEE, line 166: 🌮 nice to have the generated files to compare by hand/eye if there are differences

src/main/resources/static/css/nhsuk-9.3.0.min.css
ReubenRobertsHEE, line 1: I'm going to take it on faith that this is the same as the content in https://github.com/nhsuk/nhsuk-frontend/releases

ReubenRobertsHEE (approved):
Perfectly fine slice ;)

PR #705, opened 2025-03-27
feat(ltft): generate form ref on first submission

Judge40:
When submitted a form for the first time the form reference should be populated using the format `ltft_47165_001` where the first number is the trainee ID and the second is based on the count of previous LTFTs.

TIS21-7080
TIS21-7110


src/integrationTest/java/uk/nhs/hee/tis/trainee/forms/service/LtftServiceIntegrationTest.java
```diff
@@ -0,0 +1,159 @@
...
+import uk.nhs.hee.tis.trainee.forms.dto.LtftFormDto.StatusDto;
+import uk.nhs.hee.tis.trainee.forms.dto.LtftFormDto.StatusDto.StatusInfoDto;
+import uk.nhs.hee.tis.trainee.forms.dto.enumeration.LifecycleState;
+import uk.nhs.hee.tis.trainee.forms.dto.identity.TraineeIdentity;
+import uk.nhs.hee.tis.trainee.forms.model.LtftForm;
+
+@SpringBootTest
+@ActiveProfiles("test")
+@Testcontainers
+class LtftServiceIntegrationTest {
```
ReubenRobertsHEE, line 55: I was going to query why there isn't a test to demonstrate that form refs are not changed if a form is unsubmitted and resubmitted, but I see that scenario is covered in the service unit tests.

src/main/java/uk/nhs/hee/tis/trainee/forms/repository/LtftFormRepository.java
```diff
@@ -37,6 +37,14 @@
 @Repository
 public interface LtftFormRepository extends MongoRepository<LtftForm, UUID> {
 
+  /**
+   * Count the number of LTFT forms belonging to the given trainee.
```
ReubenRobertsHEE, line 41: That's a pretty generic description of the query :>

ReubenRobertsHEE (dismissed):

src/integrationTest/java/uk/nhs/hee/tis/trainee/forms/service/LtftServiceIntegrationTest.java
Judge40: I try and limit integration tests to things that aren't testable by unit test.

In this case the draft filtering is done with query keywords, so the _only_ way to test is via integration tests, but blocking regeneration on resubmission is coded logic which is easily covered by unit testing.

I did feel the logic for whether it was a first time submission was a little fluffy in it's intention, checking the revision might be a better approach once we have that implemented - at that point an integration test seems sensible as we'll want to verify consistent `formRef`s across two DB collections.

src/main/java/uk/nhs/hee/tis/trainee/forms/repository/LtftFormRepository.java
Judge40: Red-Green-Refactor-ReJavadoc? 😄
I've added "non-draft" to the description.

ReubenRobertsHEE (approved):
👍

PR #706, opened 2025-04-01
feat(ltft): restrict readonly properties on create

Judge40:
Several properties on the LTFT DTO should be read-only
 - `formRef`
 - `revision`
 - `assignedAdmin`
 - `status`
 - `created`
 - `lastModified`

Currently it is possible to create new LTFTs with user-provided values in these fields. Any given values should be rejected/ignored and server-generated values should be persisted and returned in the response instead.

Update LtftFormDto to add validation to ensure that the values for these properties is `null` during LTFT creation.
A `ReadOnly` JsonView should be used to make sure the deserialized DTO never contains values for these fields. This will cause the validation to always pass, but the dual-front approach ensures less chance of misconfiguration allowing values through.

TIS21-7137
TIS21-7149

### DIFF
--- a/src/integrationTest/java/uk/nhs/hee/tis/trainee/forms/api/LtftResourceIntegrationTest.java
+++ b/src/integrationTest/java/uk/nhs/hee/tis/trainee/forms/api/LtftResourceIntegrationTest.java
@@ -26,6 +26,7 @@ import static org.hamcrest.Matchers.greaterThan;
 import static org.hamcrest.Matchers.hasSize;
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.notNullValue;
+import static org.hamcrest.Matchers.nullValue;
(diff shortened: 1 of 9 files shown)


ReubenRobertsHEE (approved):
I was wondering if we should include some warning/indication in the response if we have to ignore certain submitted values, in the way that we do for errors. But it felt like low-value for the effort involved, and might require more FE changes to be useful 🤷

Judge40:
> I was wondering if we should include some warning/indication in the response if we have to ignore certain submitted values, in the way that we do for errors. But it felt like low-value for the effort involved, and might require more FE changes to be useful 🤷

I think the more sensible approach is to update the API docs to remove those fields from the request DTO. The `JsonView` approach is an alternative to splitting the DTO, but we can still conceptually treat it as two DTOs now - which would normally be documented as such in the API spec.

PR #707, opened 2025-04-01
feat: increment revision on unsubmit

ReubenRobertsHEE:
TIS21-7096

### DIFF
--- a/src/main/java/uk/nhs/hee/tis/trainee/forms/dto/enumeration/LifecycleState.java
+++ b/src/main/java/uk/nhs/hee/tis/trainee/forms/dto/enumeration/LifecycleState.java
@@ -40,6 +40,8 @@ public enum LifecycleState {
   private Set<LifecycleState> allowedTransitions;
   @Getter
   private boolean requiresDetails; //requires details to enter this state
+  @Getter
+  private boolean incrementsRevision; //increments revision no. when transitioning to this state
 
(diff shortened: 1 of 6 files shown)


src/main/java/uk/nhs/hee/tis/trainee/forms/dto/enumeration/LifecycleState.java
```diff
@@ -40,6 +40,8 @@ public enum LifecycleState {
   private Set<LifecycleState> allowedTransitions;
   @Getter
   private boolean requiresDetails; //requires details to enter this state
+  @Getter
+  private boolean incrementsRevision; //increments revision no. when transitioning to this state
```
ReubenRobertsHEE, line 44: I was wondering if this is starting to overburden the lifecycle state enum with business rules. In the end I decided it was similar enough to requiresDetails to add here, but very happy for another opinion on this.

Judge40: Given this is only ever going to be done unsubmit hardcoding it in the service would have been fine, but I'm also not too upset with this approach. Easy enough to refactor in the future if we need to simplify, and does give us an easy way to tweak the revision behaviour - which is a lot of guess work and assumption still.

Judge40 (approved):

Judge40 (approved):

Judge40 (approved):

PR #708, opened 2025-04-01
fix(ltft): restrict trainee ability to modify forms

Judge40:
fix(ltft): restrict edits to DRAFT and UNSUBMITTED
Currently the API allows edits to forms in any state, this should be
restricted to only those in DRAFT or SUBMITTED states.

TIS21-7137
TIS21-7150

---

fix(ltft): ignore read-only fields on update
When a trainee updates an LTFT it is possible to provide values for the
following fields, which should be read only:
 - `formRef`
 - `revision`
 - `assignedAdmin`
 - `status`
 - `created`
 - `lastModified`

Update LtftFormDto and LtftResource to ensure that the above fields are
either validated or excluded from deserialization using JsonViews.

Update LtftService to update the existing form content instead of
persisting the user-provided data. The provided content will be
extracted and inserted in to the existing form, with the exception of
`assignedAdmin` which will be copied from the existing form.

TIS21-7137
TIS21-7150

### DIFF
--- a/src/main/java/uk/nhs/hee/tis/trainee/forms/model/content/LtftContent.java
+++ b/src/main/java/uk/nhs/hee/tis/trainee/forms/model/content/LtftContent.java
@@ -25,6 +25,7 @@ import java.time.LocalDate;
 import java.util.List;
 import java.util.UUID;
 import lombok.Builder;
+import lombok.With;
 import org.springframework.data.mongodb.core.index.Indexed;
 import org.springframework.data.mongodb.core.mapping.Field;
 import uk.nhs.hee.tis.trainee.forms.model.Person;
@@ -50,6 +51,7 @@ public record LtftContent(
     Discussions discussions,
     CctChange change,
     Reasons reasons,
+    @With
     Person assignedAdmin) implements FormContent {
 
   /**
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -10,7 +10,7 @@ plugins {
 }
 
 group = "uk.nhs.hee.tis.trainee"
-version = "0.38.0"
+version = "0.38.1"
 
 configurations {
   compileOnly {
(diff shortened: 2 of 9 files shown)


src/main/java/uk/nhs/hee/tis/trainee/forms/model/content/LtftContent.java
```diff
@@ -50,6 +51,7 @@ public record LtftContent(
     Discussions discussions,
     CctChange change,
     Reasons reasons,
+    @With
     Person assignedAdmin) implements FormContent {
```
ReubenRobertsHEE, line 55: I'm missing something about why assignedAdmin is handled differently from the other read-only fields (annotated here, but mainly in the service). Could you add a note to explain this please 😀

ReubenRobertsHEE: Oh ok, I get it, its the one field in the LtftContent object that the trainee is not the 'master' of

Judge40: I'll add an in-line comment in the service 👍🏻

Judge40: I do keep circling round to whether the assigned admin should be lifted up to the parent form instead of in the content 🤔

ReubenRobertsHEE: Yeah, it might be neater...

Judge40: I've got a separate piece of work to add the endpoint for assigning the admin, so I'll do the refactoring there rather than adding more changes to this bug fix.

ReubenRobertsHEE (approved):

PR #709, opened 2025-04-02
feat(ltft): assign admin endpoint

Judge40:
Add an endpoint `PUT: /api/admin/ltft/{formId}/assign` to allow an admin to be assigned to the application.
Update the assigned admin and return the updated form.

TIS21-6636
TIS21-7151

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -10,7 +10,7 @@ plugins {
 }
 
 group = "uk.nhs.hee.tis.trainee"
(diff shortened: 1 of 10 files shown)


Judge40:
Handbook and local data will need updating to account for the `assignedAdmin` field being moved

ReubenRobertsHEE (approved):
Checked structures of DTO / summaries are the same, so should not break anything downstream.

PR #710, opened 2025-04-02
feat: snapshot submitted LTFT forms

ReubenRobertsHEE:
There is a rather expansive collection of indexes for the new collection, but it didn't feel worth refactoring the AbstractAuditedForm inheritance given we may in future need search functionality that depends on those indexes.

TIS21-7097


Judge40 (approved):
LGTM

PR #711, opened 2025-04-03
fix: index name length

ReubenRobertsHEE:
Service is currently failing to deploy with error
```
Failed to instantiate [org.springframework.data.mongodb.core.MongoTemplate]: Factory method 'mongoTemplate' threw exception with message: Cannot create index for 'content.programmeMembership.designatedBodyCode' in collection 'LtftSubmissionHistory' with keys 'Document{{content.programmeMembership.designatedBodyCode=1}}' and options 'Document{{name=content.programmeMembership.designatedBodyCode}}' 
...
Caused by: com.mongodb.MongoCommandException: Command failed with error 67: 'namespace name generated from index name is too long'
```

This fix assumes we are ok with manually deleting the existing index on this field in the LtftForm collection, since the name will change. Otherwise I'll cook-up a mongock.

NO-TICKET

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -10,7 +10,7 @@ plugins {
 }
 
 group = "uk.nhs.hee.tis.trainee"
-version = "0.40.0"
+version = "0.40.1"
 
 configurations {
   compileOnly {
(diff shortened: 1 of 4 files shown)


Judge40 (approved):
Nothing is "live", so I'm happy enough with manual intervention 😅

PR #713, opened 2025-04-07
fix: include empty status detail in DTO

ReubenRobertsHEE:
Tbh, I am not super happy with this, but I think we probably need a general rationalisation of what is returned null, as an empty string or an empty object or array.

Anyway, it should fix the issues the FE is having with null status details.

NO-TICKET


john12321:
> Tbh, I am not super happy with this, but I think we probably need a general rationalisation of what is returned null, as an empty string or an empty object or array.
> 
> Anyway, it should fix the issues the FE is having with null status details.
> 
> NO-TICKET

Is this on the `Not happy -> happy -> very happy -> super happy` scale? If so, I'm very happy. 😀 Never get this happy on FE.

BTW, noticed from the logs (locally) the submit error was a snsException, complaining about Topic names (whatever they are)

john12321 (approved):
LGTM (Andy may have other ideas but while the cat's away...)

ReubenRobertsHEE:
Yeah, the dev handbook needs a bit of love, we've added some stuff in the config and I think its missing from the localstack SNS/SQS

PR #714, opened 2025-04-08
feat(ltft): add supporting information

Judge40:
Add supporting information to the LTFT reasons, allowing users to provide free text to supporting their LTFT application.

TIS21-7139
TIS21-7153


DorisWongNHS (approved):
LGTM

PR #715, opened 2025-04-11
feat(ltft): capture history of assigned admins

Judge40:
Refactor the AbstractAuditedForm to move the assignedAdmin property in to the StatusInfo. This allows it to be record both against the current status and in the history.

When updating the state of a form the previous assigned admin should be carried forward, similarly when assigning an admin the previous state should be carried forward.

TIS21-7082
TIS21-7086


DorisWongNHS (approved):
LGTM

PR #718, opened 2025-04-22
feat(ltft): add managing deanery to LTFT form

Judge40:
The managing deanery/local office name is needed by the notifications service to determine which links and contact addresses to use.

TIS21-6596
TIS21-6594


DorisWongNHS (approved):
Thank you Andy!!

PR #723, opened 2025-04-24
build: migrate to version catalog

Judge40:
Use the trainee version catalog to determine plugin and dependency versions.

Update the GitHub Actions workflows to enabled build scan publishing.

NO-TICKET

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -11,6 +11,7 @@ jobs:
     uses: health-education-england/.github/.github/workflows/ci-cd-gradle.yml@main
     with:
       cluster-prefix: trainee
+      publish-build-scan: true
     secrets:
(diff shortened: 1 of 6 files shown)


Judge40:
All tests reliant on the RestControllerAdvice to translate validation
exceptions in to Problem Details began failing for unknown reasons.
It appears that the exception handler is no longer loaded/registered
during CI/CD workflow runs.

This only occurs when running on GitHub Actions, it happens even when
running the workflow against old code that was previously successful.
So far it has not been possible to reproduce this issue locally and other
trainee services, which use the same handler, are not affected.

Attempts to import or otherwise configure the exception handler have so
far failed.
The tests should be disabled and the deployed behaviour manually
verified to ensure that it is behaving properly in the real world.
Resolving the issue on the build server can follow.

EdwardBarclay (approved):

PR #728, opened 2025-05-12
fix(ltft): missing applications when paged

Judge40:
It is possible when sorting on a field with duplicate values, such as start date, for an application to be returned on multiple pages. Which causes another application to never be returned.

For example if there are 10 LTFT per page and the applications at positions 10 and 11 have the same start date, it can be random which one appears in which position in each result set. Sometimes creating "duplicates" and hiding the remaining application.

TIS21-7257


DorisWongNHS (approved):
LGTM

PR #729, opened 2025-05-20
refactor: use Json Views to restrict the admin email and name fields …

EdwardBarclay:
…serialised for trainees vs Admins for LTFT forms.

TIS21-2733


Judge40 (commented):
Is there a reason we can't use the `@JsonView` annotation on the endpoint instead of manually mapping the object in the service?

EdwardBarclay:
@Judge40 I tried to originally implement the view through the endpoints but couldn't get the view to apply correctly, Reuben came up with this method that behaved correctly so I adapted the PR to use this service based method

Judge40:
I'll have a look, I don't really like this approach when we have annotations available to us (and use them already elsewhere).

Judge40:
@EdwardBarclay I've refactored to be inline with the previous approach, I wrote some pretty comprehensive integration tests but you might want to verify via postman that both admin and trainee endpoints are returning what you expect them to.

ReubenRobertsHEE (approved):

Judge40 (dismissed):
Splitting the PersonDto seems sensible enough at this point, but I think naming the unrestricted one as `DiscussionDto` limits its reuse in the future.
I think I'd prefer something like `PersonDto` and `PersonRedactedDto` (or `RedactedPersonDto`?) rather than `DiscussionDto`.

ReubenRobertsHEE (approved):

PR #731, opened 2025-06-02
Feat/ltft email status updates

ReubenRobertsHEE:


src/main/java/uk/nhs/hee/tis/trainee/forms/dto/NotificationEventDto.java
```diff
@@ -0,0 +1,60 @@
...
+
+@Builder
+public record NotificationEventDto(
+    TisReferenceInfo tisReference,
+    @JsonAlias("type")
+    String notificationType,
+    Instant sentAt,
+    String status,
+    String statusDetail,
+    Instant lastRetry
```
ReubenRobertsHEE, line 48: At the moment. lastRetry, statusDetail and sentAt are ignored, pending finalisation of the data model for the TPD email status.

DorisWongNHS:
Thank you for putting the questions together. 

I see your concerns about the best way to store this data. Later, if we also care both the "SUBMITTED" and "APPROVED", as well as the "TPD" and "TRAINEE" notification, putting all these data in the root of the model doesn't look great for me.
For the long-term, I prefer your idea of putting ‘notificationsStatus’ into StatusInfo, so it store all the notificationsStatus of each LTFT status change throughout the whole LTFT application.

But for now, if we only care about the LTFT_SUBMITTED_TPD email, this is the simplest implementation we can quickly move forward. We can discuss and come back anytime.
I am approving the PR if you are happy to merge with this approach.

DorisWongNHS (dismissed):

ReubenRobertsHEE:
Thanks @DorisWongNHS - I appreciate you taking a look at this! If its ok with you, I'll only merge tomorrow when Naz/Andy are back from training, in case they have a clear idea of what's needed (and to avoid having to write a migrator, if it turns out there is an immediate requirement to change the data model! 😁 )

Judge40 (commented):
I keep coming back around to whether it's really an "EmailStatus" we want here, rather than something like an `EmailValidity`/`RecipientValidity`/`Validity` with values like
 - `UNKNOWN` or `UNVERIFIED` - starting value used before we send/get feedback
 - `VALID` - successfully sent, no complaints received
 - `INVALID` - sending failed
 - `WRONG_RECIPIENT` - definitely a future piece, but this could be trigger from a "if you don't believe you are the correct recipient then click here" link in the email content

src/main/java/uk/nhs/hee/tis/trainee/forms/dto/LtftFormDto.java
```diff
@@ -74,6 +75,10 @@ public record LtftFormDto(
     CctChangeDto change,
     ReasonsDto reasons,
 
+    @JsonView(ReadOnly.class)
+    @Null(groups = {Create.class, Update.class})
+    EmailValidityType tpdEmailStatus,
```
ReubenRobertsHEE, line 80: Left as 'tpdEmailStatus' to avoid breaking change for the FE; similarly, 'UNKNOWN' has been used in the enum instead of 'UNVERIFIED' to keep things consistent with current FE listing.

Judge40 (approved):
I pushed a small bit of refactoring, the rest LGTM so merge if you're happy with my addition.

PR #732, opened 2025-06-04
feat: admin ltft endpoint role constraints

ReubenRobertsHEE:
TIS21-7304

### DIFF
--- a/src/main/java/uk/nhs/hee/tis/trainee/forms/interceptor/AdminIdentityInterceptor.java
+++ b/src/main/java/uk/nhs/hee/tis/trainee/forms/interceptor/AdminIdentityInterceptor.java
@@ -42,6 +42,7 @@ public class AdminIdentityInterceptor implements HandlerInterceptor {
   private static final String GIVEN_NAME_ATTRIBUTE = "given_name";
   private static final String FAMILY_NAME_ATTRIBUTE = "family_name";
   private static final String GROUPS_ATTRIBUTE = "cognito:groups";
+  private static final String ROLES_ATTRIBUTE = "cognito:roles";
 
   private final AdminIdentity adminIdentity;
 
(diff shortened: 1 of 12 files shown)


src/main/java/uk/nhs/hee/tis/trainee/forms/interceptor/AdminIdentityInterceptor.java
```diff
@@ -60,6 +61,8 @@ public boolean preHandle(HttpServletRequest request, HttpServletResponse respons
         adminIdentity.setEmail(email);
         Set<String> adminGroups = AuthTokenUtil.getAttributes(authToken, GROUPS_ATTRIBUTE);
         adminIdentity.setGroups(adminGroups);
+        Set<String> adminRoles = AuthTokenUtil.getAttributes(authToken, ROLES_ATTRIBUTE);
+        adminIdentity.setRoles(adminRoles);
```
ReubenRobertsHEE, line 65: Added to root AdminIdentity class, though it could be limited to just the AdminLtftIdentity class, since its only used there. It felt like it made sense to include it here since it fitted with the groups and involved less code duplication in the subclass.

src/main/java/uk/nhs/hee/tis/trainee/forms/dto/identity/AdminIdentity.java
```diff
@@ -34,7 +34,8 @@ public class AdminIdentity extends UserIdentity {
 
   private static final String ROLE = "ADMIN";
 
-  private Set<String> groups;
+  protected Set<String> groups;
+  protected Set<String> roles;
```
ReubenRobertsHEE, line 38: See comment in interceptor for why this is included here

ReubenRobertsHEE:
Will try another approach.

PR #733, opened 2025-06-05
feat: limit admin/ltft endpoints

ReubenRobertsHEE:
Based on cognito role. Same approach as used in tis-trainee-details cct endpoints, fwiw.

TIS21-7304


Judge40 (approved):
Nice.

PR #735, opened 2025-06-06
feat: limit trainee ltft endpoints to rollout PMs

ReubenRobertsHEE:
Checks on creating/updating a LTFT.

Note, no special status / message is returned from the API calls if they fail, just the generic 'Bad Request'. If we need this, should it be in a follow-up ticket, or added as an A/C?

TIS21-7305

### DIFF
--- a/src/integrationTest/java/uk/nhs/hee/tis/trainee/forms/TestJwtUtil.java
+++ b/src/integrationTest/java/uk/nhs/hee/tis/trainee/forms/TestJwtUtil.java
@@ -24,6 +24,7 @@ package uk.nhs.hee.tis.trainee.forms;
 import java.nio.charset.StandardCharsets;
 import java.util.Base64;
 import java.util.List;
+import java.util.UUID;
 import java.util.stream.Collectors;
 
 /**
@@ -36,7 +37,7 @@ public class TestJwtUtil {
   public static final String GIVEN_NAME_ATTRIBUTE = "given_name";
   public static final String FAMILY_NAME_ATTRIBUTE = "family_name";
   public static final String FEATURES_ATTRIBUTE = "features";
-  public static final String FEATURES_LTFT_PROGRAMME = "607ddcfd-bbe9-4835-b795-ca26aebba990";
+  public static final UUID FEATURES_LTFT_PROGRAMME_INCLUDED = UUID.randomUUID();
 
   /**
    * Generate a token with the given payload.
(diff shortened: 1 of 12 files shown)


Judge40 (changes requested):
Considering the allowed programmes are going to be added to the token, we could get away with checking the linked programme ID against the token's list and avoid the service-to-service call.

src/integrationTest/java/uk/nhs/hee/tis/trainee/forms/TestJwtUtil.java
```diff
@@ -70,15 +61,28 @@ public static String generateTokenForTisId(String traineeTisId) {
...
     return generateToken(payload);
   }
 
+  /**
+   * Generate a token with default values for the various attributes as the payload.
+   *
+   * @param traineeTisId The TIS ID to inject in to the payload.
+   * @return The generated token.
+   */
+  public static String generateTokenForTrainee(String traineeTisId) {
```
ReubenRobertsHEE, line 82: This is a bit clunky, but replaces the 'generateTokenForTisId()' that various tests were using, to ensure that missing features don't either break tests or mask negative tests.

src/integrationTest/java/uk/nhs/hee/tis/trainee/forms/api/LtftResourceIntegrationTest.java
```diff
@@ -654,10 +740,14 @@ void shouldIgnoreReadOnlyFieldsWhenUpdatingLtftFormForTrainee() throws Exception
             .build())
         .created(Instant.EPOCH)
         .lastModified(Instant.EPOCH)
+        .programmeMembership(LtftFormDto.ProgrammeMembershipDto.builder()
+            .id(PM_UUID)
+            .build())
         .build();
 
+    sleep(10); // Ensure lastModified is different from created
```
ReubenRobertsHEE, line 748: Blazing fast laptop, what can I say 😅

Judge40 (dismissed):

Judge40:
> > Considering the allowed programmes are going to be added to the token, we could get away with checking the linked programme ID against the token's list and avoid the service-to-service call.
> 
> I've refactored, but sadly the no. of lines has only bloated somewhat. It's the new need for programme membership data in the LTFT form that has swelled the tests, mostly.

I'll take that over inter-service API calls all day long 😄

PR #738, opened 2025-06-26
feat: add message attributes to SNS publishing

ReubenRobertsHEE:
to allow us to distinguish between TPD email status updates and real form updates using an SNS subscription filter (upcoming PR).

TIS21-7443

### DIFF
--- a/src/main/java/uk/nhs/hee/tis/trainee/forms/service/EventBroadcastService.java
+++ b/src/main/java/uk/nhs/hee/tis/trainee/forms/service/EventBroadcastService.java
@@ -43,6 +43,7 @@ import uk.nhs.hee.tis.trainee.forms.dto.LtftFormDto;
 public class EventBroadcastService {
 
   protected static final String MESSAGE_ATTRIBUTE_KEY = "trigger";
+  protected static final String MESSAGE_ATTRIBUTE_DEFAULT_VALUE = "default";
 
   private final SnsClient snsClient;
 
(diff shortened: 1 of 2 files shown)


src/main/java/uk/nhs/hee/tis/trainee/forms/service/EventBroadcastService.java
```diff
@@ -0,0 +1,121 @@
...
+    if (snsTopic == null || snsTopic.isBlank()) {
+      log.warn("SNS topic ARN is null or blank, skipping SNS publish.");
+      return null;
+    }
+
+    PublishRequest.Builder request = PublishRequest.builder()
+        .message(eventJson.toString())
+        .topicArn(snsTopic);
+
+    if (messageAttribute != null) {
```
Judge40, line 108: Given previous experiences, I think we need to set a default value (or always require one) or else the message will get DLQ'ed

src/main/java/uk/nhs/hee/tis/trainee/forms/service/EventBroadcastService.java
```diff
@@ -0,0 +1,121 @@
...
+
+    PublishRequest request = buildSnsRequest(snsTopic, eventJson, messageAttribute, formDto.id());
+
+    if (request != null) {
+      try {
+        snsClient.publish(request);
+        log.info("Broadcast event sent to SNS for LTFT form {} with attribute {}.",
+            formDto.id(), messageAttribute);
+      } catch (SnsException e) {
+        String message = String.format(
```
Judge40, line 80: If we catch this without rethrowing _something_, then any triggering message will be deleted from the queue instead of retrying. I don't think that's what we want.

ReubenRobertsHEE: Good catch.

src/main/java/uk/nhs/hee/tis/trainee/forms/service/EventBroadcastService.java
ReubenRobertsHEE: As it happens, there is always an attribute included, but you're right: it is a gotcha. I've added a default.

PR #746, opened 2025-09-02
feat: formR PDFs

ReubenRobertsHEE:
Requires API Gateway additions before FE work: https://github.com/Health-Education-England/TIS-OPS/pull/1052

Note: FE must include programmeName in requests.

The PR is a bit repetitive because no refactoring of FormR handling has been done so far, so code is largely duplicated. Some initial work on that was done here (https://github.com/Health-Education-England/tis-trainee-forms/tree/spike/refactorFormR), but it was drifting towards mirroring the AbstractForm / AbstractFormR model but in the DTOs, and I understand inheritance in DTOs can be a bit of an anti-pattern, so maybe something that should be discussed before progressing any further.

TIS21-6583

### DIFF
--- a/src/main/java/uk/nhs/hee/tis/trainee/forms/model/AbstractFormR.java
+++ b/src/main/java/uk/nhs/hee/tis/trainee/forms/model/AbstractFormR.java
@@ -22,6 +22,7 @@
 package uk.nhs.hee.tis.trainee.forms.model;
 
 import java.time.LocalDateTime;
+import java.util.UUID;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
 import lombok.Setter;
@@ -37,5 +38,6 @@ public abstract class AbstractFormR extends AbstractForm {
   private LifecycleState lifecycleState;
   private LocalDateTime submissionDate;
   private LocalDateTime lastModifiedDate;
+  private UUID programmeMembershipId;
 
 }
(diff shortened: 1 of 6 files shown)


src/main/java/uk/nhs/hee/tis/trainee/forms/api/FormRPartBResource.java
```diff
@@ -181,4 +190,37 @@ public ResponseEntity<Void> deleteFormRPartBById(@PathVariable String id) {
...
   }
+
+  /**
+   * Generate a FormR PartB PDF, unless one already exists.
+   *
+   * @return The downloaded or generated PDF.
+   * @throws IOException A new PDF could not be generated.
+   */
+  @PutMapping(value = "/formr-partb-pdf", produces = MediaType.APPLICATION_PDF_VALUE)
+  public ResponseEntity<byte[]> generatePdf(@Valid @RequestBody FormRPartBDto formRPartB)
```
Judge40, line 201: I should have picked this up earlier, but do we want to trust the incoming DTO?

The alternative would be similar to [LTFT](https://github.com/Health-Education-England/tis-trainee-forms/blob/04b6589250c954b6e3dd862167e4c01fc21f7a90/src/main/java/uk/nhs/hee/tis/trainee/forms/api/LtftResource.java#L106) where we have two identical `GET` mappings that are selected based on the accepted type header.

Are we taking the DTO as parameter in this case because we may not have the latest data available to us (either the S3/DB split or because they're printing an updated copy they haven't saved)?

ReubenRobertsHEE: (Sorry, my latest reply was without having seen your prior comment. Dratted page-refresh issue, I often have it)

PR #749, opened 2025-09-19
Feat: formR PDF when submitted

ReubenRobertsHEE:
This enables PDFs for submitted FormRs to be published, in the same way that CoJ PDF's are published.

They will be picked up by SNS (tis-trainee-[env]-pdf-generated.fifo), and a new subscription has been added (tis-trainee-notifications-[env]-formr-pdf-generated.fifo), filtered to only pick up formR form_types. 

A new listener will be added to the notifications service to generate the emails and attach the PDFs, and the existing listener for general saved form Rs that generates the generic email notification will be adjusted to ignore those that are SUBMITTED, to avoid duplication of emails for submitted formRs, but to allow any other form states to still generate these generic emails if the 'alwaysStoreFiles' config is ever set to true in future (cf. https://github.com/Health-Education-England/tis-trainee-forms/blob/9ec182c3ec96d6a88439edcf74ce2cc8a69ab477/src/main/java/uk/nhs/hee/tis/trainee/forms/service/FormRPartAService.java#L99).

TIS21-6582

### DIFF
--- a/src/main/java/uk/nhs/hee/tis/trainee/forms/api/FormRPartAResource.java
+++ b/src/main/java/uk/nhs/hee/tis/trainee/forms/api/FormRPartAResource.java
@@ -22,7 +22,10 @@
 package uk.nhs.hee.tis.trainee.forms.api;
 
 import com.amazonaws.xray.spring.aop.XRayEnabled;
+import io.awspring.cloud.s3.Location;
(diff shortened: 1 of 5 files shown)


DorisWongNHS (approved):
LGTM

PR #817, opened 2026-04-22
feat: S3 forms no longer included in GET requests

Judge40:
When a trainee's forms are retrieved the S3 and DB forms are combined, now that S3 is no longer required for admin access to forms the GET requests should be answered fully from the database.

The current set-up has caused some issues with historical data, where unsubmitted forms were never updated in the database. By cutting S3 out, we can ensure that the trainee has a consistent experience. However, some forms may appear to "rollback" to the unsubmitted state - but this only applies where a form was unsubmitted and never updated afterwards.

NO-TICKET

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -10,7 +10,7 @@ plugins {
 }
 
 group = "uk.nhs.hee.tis.trainee"
-version = "0.65.5"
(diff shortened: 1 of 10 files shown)


DorisWongNHS (approved):
LGTM

PR #823, opened 2026-05-26
fix: remove CctDate from LTFT form

DorisWongNHS:
Sample LTFT change summary in PDF for trainees:
image

TIS21-8666
TIS21-8633


Judge40 (commented):
I think a migrator to remove existing `cctDate` data may be appropriate, up to you whether to do that here or as a followup.

EdwardBarclay (approved):
LGTM

PR #825, opened 2026-06-01
fix: stdise name of LTFT altStartDate

ReubenRobertsHEE:
To avoid alternate/alternative/etc gotchas

No need to worry about real data migration, since it was never made live (or tested on staging, for that matter)

NO-TICKET


john12321 (approved):
As an alternative to alternate it is much apreci...appreci..welcomed.

PR #836, opened 2026-07-10
feat: migrate failed Form R Part B

Judge40:
Several Form R Part B failed migration due to an unexpected state, with no record of their first submission available in S3.

Create a migrator to deal with these failures, it will assume the following due to lack of other available data.
 - All forms are the trainee's first submission (form ref 001)
 - The forms had a single submission (revision data lost)
 - There is no reliable history in S3
 - All forms are at SUBMITTED in the DB

As a result a lot of assumptions can be made and the original migrator can be trimmed down significantly.
 - Only two history items: DRAFT -> SUBMITTED
 - Form Ref is `formr_partb_{traineeId}_001`
 - ModifiedBy is always the trainee themselves
 - Timestamps will all copy the known submission timestamp

TIS21-8706

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -10,7 +10,7 @@ plugins {
 }
 
 group = "uk.nhs.hee.tis.trainee"
-version = "0.69.1"
+version = "0.70.0"
 
 configurations {
   compileOnly {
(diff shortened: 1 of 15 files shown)


ReubenRobertsHEE (approved):
I haven't picked-through the tests in huge detail; it feels like you deserve that weekend to start and we can validate on staging 😁

PR #837, opened 2026-07-21
fix: duplicate FormR formRefs

Judge40:
During the FormR migration several forms were submitted by trainees while their existing forms were still being processed. This led to some duplicate form references being used.

The formRef does not yet have a unique index, but this will be added in a follow-up once the migrator has fixed the existing duplications.

TIS21-8867
TIS21-8868

### DIFF
--- a/src/main/java/uk/nhs/hee/tis/trainee/forms/migration/FixDuplicateFormrRefs.java
+++ b/src/main/java/uk/nhs/hee/tis/trainee/forms/migration/FixDuplicateFormrRefs.java
@@ -1,0 +1,230 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright 2026 Crown Copyright (Health Education England)
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of this software and
+ * associated documentation files (the "Software"), to deal in the Software without restriction,
+ * including without limitation the rights to use, copy, modify, merge, publish, distribute,
+ * sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all copies or
+ * substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT
+ * NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+ * NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+ * DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ *
+ */
+
+package uk.nhs.hee.tis.trainee.forms.migration;
+
+import io.mongock.api.annotations.ChangeUnit;
+import io.mongock.api.annotations.Execution;
+import io.mongock.api.annotations.RollbackExecution;
+import java.util.List;
+import java.util.UUID;
+import lombok.extern.slf4j.Slf4j;
+import org.bson.Document;
+import org.springframework.data.domain.Sort;
+import org.springframework.data.mongodb.core.MongoTemplate;
+import org.springframework.data.mongodb.core.aggregation.Aggregation;
+import org.springframework.data.mongodb.core.query.Criteria;
+import org.springframework.data.mongodb.core.query.Query;
+import uk.nhs.hee.tis.trainee.forms.model.AbstractFormR;
+import uk.nhs.hee.tis.trainee.forms.model.FormRPartA;
+import uk.nhs.hee.tis.trainee.forms.model.FormRPartB;
+import uk.nhs.hee.tis.trainee.forms.model.FormSubmissionHistory;
+import uk.nhs.hee.tis.trainee.forms.model.FormrPartaSubmissionHistory;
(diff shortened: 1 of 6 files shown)


src/main/java/uk/nhs/hee/tis/trainee/forms/migration/FixDuplicateFormrRefs.java
```diff
@@ -0,0 +1,230 @@
...
+    final int formCount = forms.size();
+    int fixCount = 0;
+
+    for (int i = 0; i < formCount; i++) {
+      Document form = forms.get(i);
+      String formRef = form.getString(FIELD_FORM_REF);
+      String formRefPrefix = formRef.substring(0, formRef.length() - FORM_REF_SUFFIX_LENGTH);
+      int formRefSuffix = Integer.parseInt(
+          formRef.substring(formRef.length() - FORM_REF_SUFFIX_LENGTH));
+      int expectedSuffix = formCount - i; // Forms are latest first.
```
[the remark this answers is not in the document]
Judge40: All known data conforms

src/main/java/uk/nhs/hee/tis/trainee/forms/service/FormRPartAService.java
```diff
@@ -342,7 +342,7 @@ public void publishUpdateNotification(FormRPartADto form, String snsTopic) {
    *
    * @param formDto The form DTO to publish.
    */
-  private void publishFormRUpdateEvent(FormRPartADto formDto) {
+  public void publishFormRUpdateEvent(FormRPartADto formDto) {
```
Judge40, line 345: Not ideal to change the visibility, but avoids overcomplicating the migrator and the update notifications will be reworked in the near future anyway.

src/main/java/uk/nhs/hee/tis/trainee/forms/service/FormRPartBService.java
```diff
@@ -343,7 +343,7 @@ public void publishUpdateNotification(FormRPartBDto form, String snsTopic) {
    *
    * @param formDto The form DTO to publish.
    */
-  private void publishFormRUpdateEvent(FormRPartBDto formDto) {
+  public void publishFormRUpdateEvent(FormRPartBDto formDto) {
```
Judge40, line 346: Not ideal to change the visibility, but avoids overcomplicating the migrator and the update notifications will be reworked in the near future anyway.

src/main/java/uk/nhs/hee/tis/trainee/forms/migration/FixDuplicateFormrRefs.java
```diff
@@ -0,0 +1,230 @@
...
+    this.partaService = partaService;
+    this.partbService = partbService;
+  }
+
+  /**
+   * Recalculate form references for duplicated forms.
+   */
+  @Execution
+  public void migrateCollections() {
+    int fixCount = migrateCollection(FormRPartA.class, FormrPartaSubmissionHistory.class);
```
JosephKelly, line 82: 260 is far short of `Integer.MAX_VALUE` so it seems like a good choice, expecting this won't be run again on a significantly larger data store.

src/main/java/uk/nhs/hee/tis/trainee/forms/migration/FixDuplicateFormrRefs.java
```diff
@@ -0,0 +1,230 @@
...
+
+    List<String> traineeIds = getTraineeIdsWithDuplicates(formCollection);
+
+    log.info("Found {} trainee(s) with duplicate formRefs in collection {}.", traineeIds.size(),
+        formCollection);
+
+    int fixCount = traineeIds.stream()
+        .mapToInt(
+            traineeId -> fixTraineeForms(traineeId, formClass, historyClass, formCollection,
+                historyCollection))
```
JosephKelly, line 100: Acceptable now but additional parameters add to cognitive load.  I can see the `fromCollection` is logged outside the method where it's used.

You could remove at least the `historyCollection` and if `formClass.getSimpleName()` was a good enough proxy for `fromCollection` in logging then too.

src/main/java/uk/nhs/hee/tis/trainee/forms/migration/FixDuplicateFormrRefs.java
```diff
@@ -0,0 +1,230 @@
...
+  private void publishFormUpdate(Document form, Class<? extends AbstractFormR<?>> formClass) {
+    if (formClass.equals(FormRPartA.class)) {
+      log.debug("Publishing FormRPartA update event for new form ref {}",
+          form.getString(FIELD_FORM_REF));
+      partaService.getAdminsFormRPartAById(form.get(FIELD_ID, UUID.class).toString()).ifPresent(
+          partaService::publishFormRUpdateEvent);
+    } else {
+      log.debug("Publishing FormRPartB update event for new form ref {}",
+          form.getString(FIELD_FORM_REF));
+      partbService.getAdminsFormRPartBById(form.get(FIELD_ID, UUID.class).toString()).ifPresent(
```
JosephKelly, line 217: I can see that the `publishFormRUpdateEvent` can't reasonably be replaced and are good and stable.  I expect the `.getAdmins...()` methods are very stable and unlikely to change too.  They look like they could potentially be replaced by direct queries to further improve on decoupling from the application code.

Really interesting to see this use of a change set that follows guidance on being independent of application code like MongoRepositories!

JosephKelly (approved):
Within a brief look thinking of java standards a quick pass looks good, recognising this sounds time-sensitive.

There's some variation is spacing which is likely acceptable for readability and with a more thorough review, I'd probably look at the operations to check if they could be simplified.

src/main/java/uk/nhs/hee/tis/trainee/forms/migration/FixDuplicateFormrRefs.java
Judge40: There was reason for not using direct queries here too, by using the service methods I can get by ID and pass the DTO without having to worry about it's shape too much. Otherwise, to transform the Document to then DTO I have to use Jackson to convert to the entity, then the mapstruct mapper to convert to the DTO, which felt much clunkier.

It does reflect the more general need to improve how we emit this type of update event (application events are a future spike).

PR #838, opened 2026-07-22
feat: make `formRef` unique for audited forms

Judge40:
The `formRef` field should always be unique but there is no enforcement of it.
Update the form classes to change the `formRef` index to ensure uniqueness, sparse should be used to ignore non-submitted forms which won't have a formRef.

Create a `CurrentForm` to act as live form equivalent of `FormSubmissionHistory`, each of Form R A/B and LTFT will implement this interface to inherit the indexing.

Move the `formRef` indexing from `AbstractAuditedForm` to `CurrentForm` and `FormSubmissionHistory` to allow differing index types between the actual form and the history snapshots.

As Mongock is unable to replace the existing indexes without disabling auto-index creation, the existing indexes must be manually deleted to allow deployment.

TIS21-8867
TIS21-8869

### DIFF
--- a/src/main/java/uk/nhs/hee/tis/trainee/forms/model/CurrentForm.java
+++ b/src/main/java/uk/nhs/hee/tis/trainee/forms/model/CurrentForm.java
@@ -22,29 +22,12 @@
 
 package uk.nhs.hee.tis.trainee.forms.model;
 
-import lombok.Data;
-import lombok.EqualsAndHashCode;
-import lombok.ToString;
-import org.springframework.data.mongodb.core.mapping.Document;
-import uk.nhs.hee.tis.trainee.forms.model.content.FormrPartaContent;
+import org.springframework.data.mongodb.core.index.CompoundIndex;
 
 /**
- * A FormR Part A form submission entity.
+ * An interface representing the current version of a form, stored in the main collection.
  */
-@Document("FormrPartaSubmissionHistory")
-@Data
-@ToString(callSuper = true)
-@EqualsAndHashCode(callSuper = true)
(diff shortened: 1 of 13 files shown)


src/main/java/uk/nhs/hee/tis/trainee/forms/model/CurrentForm.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package uk.nhs.hee.tis.trainee.forms.model;
+
+import org.springframework.data.mongodb.core.index.CompoundIndex;
+
+/**
+ * An interface representing the current version of a form, stored in the main collection.
+ */
+@CompoundIndex(name = "formRef", def = "{'formRef': 1}", unique = true, sparse = true)
```
ReubenRobertsHEE, line 30: sparse means the nulls for DRAFT forms work ok ✅

ReubenRobertsHEE (dismissed):

Judge40:
@ReubenRobertsHEE It's really awkward because it's failing to set up the test scenario with duplicate formRefs because they can't be duplicated anymore!

I have the choice of deleting the index as part of the test set up, or we decide what to do with these redundant/broken migrations long term.

See #839 for a similar issue, I had to refactor a couple of other migrations to make them "work" after adding required auth to service calls and removing S3. Those migrations will also break when #838 is merged as we do insert > delete which would temporarily duplicate the formRef 🥴

Judge40:
This change breaks several historical DB migrations, a decision will need to be made about how we manage these situations going forward. Until that decision is made this PR is on hold and the original sub-task has been converted to a story in the backlog.

PR #839, opened 2026-07-27
refactor: form-r ID migration for UUID

Judge40:
There are two migrations ran on FormR to migrate from ObjectId to UUID for the ID field.
Since the Form-R refactoring these required some refactoring to bring them back in line with up-to-date expectations.

 1. Remove the S3 updates as S3 form mirroring has been removed.
 2. Switch service usage to raw Document to avoid the migration applying a `modifiedBy` to each form, which requires an unavailable Identity/claims.

NO-TICKET

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -10,7 +10,7 @@ plugins {
 }
 
 group = "uk.nhs.hee.tis.trainee"
-version = "0.70.1"
+version = "0.71.2"
 
 configurations {
   compileOnly {
--- a/src/integrationTest/java/uk/nhs/hee/tis/trainee/forms/api/FormRPartAResourceIntegrationTest.java
+++ b/src/integrationTest/java/uk/nhs/hee/tis/trainee/forms/api/FormRPartAResourceIntegrationTest.java
@@ -425,7 +425,7 @@ class FormRPartAResourceIntegrationTest {
         .andExpect(status().isCreated())
         .andExpect(jsonPath("$.lifecycleState").value("SUBMITTED"));
 
-    ArgumentCaptor<PublishRequest> requestCaptor = ArgumentCaptor.forClass(PublishRequest.class);
+    ArgumentCaptor<PublishRequest> requestCaptor = ArgumentCaptor.captor();
     verify(snsClient, times(2)).publish(requestCaptor.capture());
 
     PublishRequest publishRequest1 = requestCaptor.getAllValues().get(0);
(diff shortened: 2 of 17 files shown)


ReubenRobertsHEE:
Is there a reason why we don't simply stub these migrations, to avoid having the ongoing maintenance burden? 

To me, it seems vanishingly unlikely that anyone else is currently using the codebase and would need to apply these migrations in future (and I would think the retroactive changes to the migrators would cause issues for them anyway) - if we wanted to support that, wouldn't tagged releases be the way to go? Or am I missing something? 😀

Judge40:
> Is there a reason why we don't simply stub these migrations, to avoid having the ongoing maintenance burden?
> 
> To me, it seems vanishingly unlikely that anyone else is currently using the codebase and would need to apply these migrations in future (and I would think the retroactive changes to the migrators would cause issues for them anyway) - if we wanted to support that, wouldn't tagged releases be the way to go? Or am I missing something? 😀

That's pretty much my preferred option, I think throwing an exception if the migrator is ran, alongside a documented migration path covers both our own needs and at least feels like we're following best practice for any hypothetical consumer of our code.

What will need a bit more attention is how we write future proof migrations, arguably all the issues I've encountered were a result of bad practice and highlights a need to define standards for our migrators in the future.
Things like calling service methods instead of doing the work generically on the DB documents, or loading the whole collection in to memory instead of paging. As the service has evolved and the DB has grown we've found ourselves unable to even run quite a few of the older migrations - which I never noticed until I tried to start up a fresh dev env.

ReubenRobertsHEE:
> > Is there a reason why we don't simply stub these migrations, to avoid having the ongoing maintenance burden?
> > To me, it seems vanishingly unlikely that anyone else is currently using the codebase and would need to apply these migrations in future (and I would think the retroactive changes to the migrators would cause issues for them anyway) - if we wanted to support that, wouldn't tagged releases be the way to go? Or am I missing something? 😀
> 
> That's pretty much my preferred option, I think throwing an exception if the migrator is ran, alongside a documented migration path covers both our own needs and at least feels like we're following best practice for any hypothetical consumer of our code.
> 
> What will need a bit more attention is how we write future proof migrations, arguably all the issues I've encountered were a result of bad practice and highlights a need to define standards for our migrators in the future. Things like calling service methods instead of doing the work generically on the DB documents, or loading the whole collection in to memory instead of paging. As the service has evolved and the DB has grown we've found ourselves unable to even run quite a few of the older migrations - which I never noticed until I tried to start up a fresh dev env.

Do you want to merge this PR and then move in that direction (and set up some standards for how we should write migrators in future), or were you thinking to stub it right now?

Judge40:
> Do you want to merge this PR and then move in that direction (and set up some standards for how we should write migrators in future), or were you thinking to stub it right now?

I think we merge this one now, our dev env data is actually set up to need these two migrators so it fixes that problem at least.

ReubenRobertsHEE (approved):

PR #842, opened 2026-07-30
fix: remove AGENTS.md

ReubenRobertsHEE:
Not required (at this point)

NO-TICKET


ReubenRobertsHEE:
@DorisWongNHS would you mind looking at this? It slipped through in the last PR, but we haven't talked as a team about whether to set these up, so probably best to remove for now!

Judge40 (approved):

PR #853, opened 2026-08-14
fix: set environment properly

ReubenRobertsHEE:
To fix the manual setting hack of SPRING_PROFILES_ACTIVE = preprod to make the review stages testable on preprod.

NO-TICKET

### DIFF
--- a/.aws/task-definition-template.json
+++ b/.aws/task-definition-template.json
@@ -79,6 +79,10 @@
         {
           "name": "SIGNATURE_SECRET_KEY",
           "valueFrom": "/tis/trainee/${environment}/signature/secret-key"
+        },
+        {
+          "name": "SPRING_PROFILES_ACTIVE",
+          "valueFrom": "/tis/trainee/forms/${environment}/spring-profiles-active"
         }
       ],
       "logConfiguration": {
--- a/.aws/task-definition-template.json
+++ b/.aws/task-definition-template.json
@@ -79,10 +79,6 @@
         {
           "name": "SIGNATURE_SECRET_KEY",
           "valueFrom": "/tis/trainee/${environment}/signature/secret-key"
-        },
-        {
-          "name": "SPRING_PROFILES_ACTIVE",
-          "valueFrom": "/tis/trainee/forms/${environment}/spring-profiles-active"
         }
       ],
       "logConfiguration": {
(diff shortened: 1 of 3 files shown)


EdwardBarclay (dismissed):

Judge40 (changes requested):
I think I'd prefer setting `SPRING_PROFILES_ACTIVE` in the task definition, since this is more of a deployment concern than an app configuration concern.

I'm definitely a bit dubious of `local` as default, that's potentially risky on prod in that it would cover up missing required configuration etc. by forcing the local defaults (which would still fail, but in a less obvious way).

In general, I want us to move away from "local" style defaults in the main `application.yml` and move it all to `application-local.yml` now we tend to have that in most repositories.
The application absolutely should fail to start if we've failed to provide the correct env vars during deployment.

ReubenRobertsHEE:
Good shout @Judge40 . I have separated the setting out.

Thinking to populate it as follows (i.e. not using {$local.environment}, to avoid any unintended coupling): does this look ok?

```

resource "aws_ssm_parameter" "trainee_forms_preprod_spring_profiles_active" {
  name  = "/tis/trainee/forms/preprod/spring-profiles-active"
  type  = "String"
  value = "preprod"
}

resource "aws_ssm_parameter" "trainee_forms_prod_spring_profiles_active" {
  name  = "/tis/trainee/forms/prod/spring-profiles-active"
  type  = "String"
  value = "prod"
}
```

.aws/task-definition-template.json
```diff
@@ -79,6 +79,10 @@
         {
           "name": "SIGNATURE_SECRET_KEY",
           "valueFrom": "/tis/trainee/${environment}/signature/secret-key"
+        },
+        {
+          "name": "SPRING_PROFILES_ACTIVE",
+          "valueFrom": "/tis/trainee/forms/${environment}/spring-profiles-active"
```
Judge40: We could just use `${environment}` and move it to the environment section instead of secrets I think?

Judge40 (changes requested):

.aws/task-definition-template.json
ReubenRobertsHEE: Ah ok, I thought keeping app config and deployment concerns separate might need a separate value, but if not, that is simpler

Judge40 (approved):

.aws/task-definition-template.json
Judge40: It's fine for now since we don't need layered profiles, if we start having profiles for metrics etc. then we may need to change in the future.

